PR #4, opened 2018-12-24
Add afghanistan_releases crawler

BibianaC:
Kingfisher crawler:
https://github.com/open-contracting/kingfisher/blob/master/ocdskingfisher/sources/afghanistan_releases.py

### DIFF
--- a/kingfisher_scrapy/spiders/afghanistan_releases.py
+++ b/kingfisher_scrapy/spiders/afghanistan_releases.py
@@ -1,0 +1,18 @@
+import json
+
+import requests
+import scrapy
+
+
+class AfghanistanReleases(scrapy.Spider):
+    name = 'afghanistan_releases'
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/spiders/afghanistan_releases.py
```diff
@@ -0,0 +1,18 @@
...
+import scrapy
+
+
+class AfghanistanReleases(scrapy.Spider):
+    name = 'afghanistan_releases'
+    start_urls = ['https://ocds.ageops.net/api/ocds/releases/dates']
+    download_delay = 1
+
+    def parse(self, response):
+        for url in json.loads(response.body):
```
BibianaC: @rhiaro would you 'yield' them by blocks or all together?

rhiaro: Doing them one by one might let scrapy schedule them better, but I'm not 100% sure. See which is faster?

PR #7, opened 2019-01-10
Spider fixes

rhiaro:
Emit file URLs one by one instead of in chunks so requests are none-blocking


kingfisher_scrapy/spiders/afghanistan_releases.py
```diff
@@ -11,16 +11,25 @@ class AfghanistanReleases(scrapy.Spider):
     def parse(self, response):
         urls = json.loads(response.body)
 
-        if self.sample:
+        print(urls)
```
BibianaC: does this print need to be deleted?

rhiaro: I did it AGAIN

PR #8, opened 2019-01-11
Use requests for API post, log failures.

rhiaro:
This PR goes back to using `requests` for the API posting. 

This is a stop-gap to meet the immediate need of getting data into the database ASAP in the crawling process, as with using Scrapy's' inbuilt Request the POSTs get added to the queue and don't start until about half way through the crawl.

We need more work to figure out how to change how Scrapy's scheduler prioritises requests to fix this for the long term.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 scrapy >= 1.5.1
+requests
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
-scrapy >= 1.5.1
\ No newline at end of file
+scrapy >= 1.5.1
+requests
```
BibianaC: is it possible to add the version?

rhiaro: Updated

PR #10, opened 2019-01-15
Add google cloud

BibianaC:


BibianaC:
Thanks @rhiaro, very good comments.

1. I have renamed requirements.
2. I have fixed it. Now it can be used in both.
3. How have you seen that GCS URI has the month and the year.

This morning we ran a crawler in scrapyhub and we could not see in kingfirsher db the dates for that one changing.

rhiaro:
@BibianaC The `FILES_STORE` config is set to `'gs://kingfisher-scrape-jan-2019/'` Is that temporary, or do we need to change it every month, or something else?

> This morning we ran a crawler in scrapyhub and we could not see in kingfirsher db the dates for that one changing.

I don't know what this means. But I was referring to the output of the spiders being stored in `<FILES_STORE>/<spider_name>/<start_datetime>/` - the KingfisherFilesPipeline does that (https://github.com/open-contracting/kingfisher-scrape/blob/add-google-cloud/kingfisher_scrapy/pipelines.py#L36) and as far as I can tell that should still be carried out when data is saved to GCS as well? Doesn't affect the database.

BibianaC:
@rhiaro Thanks for the tip, it now saves files in folders.

PR #14, opened 2019-01-18
Fix filename issue

yolile:
If a url has a . after the last / and the text after the . is
not a valid extension, the file is saved with the incorrect
file extension. This fix use the content type of the response
to also save the file as json if the content type it is.

For example:

- http://url.json?limit=1 is now saved as .json?limit=1 , with the fix is
saved as .json


odscjames:
I unfortunately have to close this - a new pull request has to made against the master branch now.

PR #23, opened 2019-01-23
Add Armenia scraper

yolile:
was #11


jpmckinney:
Re-opening #11 instead.

PR #12, opened 2019-01-18
Australia NSW scraper

yolile:

PR #24, opened 2019-01-23
Australia nsw scraper

yolile:
was #12


odscjames:
This has conflicts with base now, due to the changes to kingfisher_scrapy/pipelines.py - but they just look like whitespace changes? (I'll set up Flake8 soon to avoid problems like this!)

Can you submit this again, with just the kingfisher_scrapy/spiders/australia_nsw.py file?

yolile:
done!

PR #27, opened 2019-01-23
More docs

odscjames:


odscjames:
> The kingfisher-process part is going to be the same for every environment you run the spiders in, and it's always optional, so maybe we don't need to repeat this?

It's not, it needs to be configured differently. Note in one of them it says the env variables need to be configured in the local environment, and the other in the scrapyd enviroment. I did wonder if I should make the later one clearer.

PR #36, opened 2019-01-25
Change to new format the API expects

odscjames:


kingfisher_scrapy/pipelines.py
```diff
@@ -106,11 +108,28 @@ def _build_api_url(crawler):
...
+            is_sample = True
+        else:
+            is_sample = False
+        data_version = self._get_start_time(spider).strftime("%Y-%m-%d %H:%M:%S")
         for completed in item:
 
-            headers['Content-Type'] = 'application/json'
-
-            response = requests.post(url, json=completed, headers=headers)
+            data = {
```
rhiaro: This is all set in the item_completed stage in the FilesPipeline and passed to here (`item` is a list of these dicts for each file downloaded - which should normally be one, but it depends on the spider). We only need to set it in one place, and because of the chance of one request downloading multiple files, it should be in `item_completed` I think.

odscjames: I think with Google Cloud that might have to change. I think we'll want to have 2 files pipeline, and so the whole question of where the dict is build get murkier. i thought it would be simpler if as much as possible, the dict was build where it was neeeded.

odscjames: Also means any dict we pass through pipeline and pass to api is different and that might be good. eg the filename is currently sent as data/canada_buyandsell/20190125_104536/6ca156064bc49ced9b208b365833a71bc843ff9f.json and it should really be   6ca156064bc49ced9b208b365833a71bc843ff9f.json

PR #38, opened 2019-01-25
Remove Scrapy Cloud Support

odscjames:
Closes https://github.com/open-contracting/kingfisher-scrape/issues/33
Closes https://github.com/open-contracting/kingfisher-scrape/issues/32


odscjames:
It's better not to have code that we aren't using in the repo; it will get stale and we might be asked support questions about it. It's in the git history if we want to go back to it later.

PR #54, opened 2019-01-30
Change KINGFISHER_API_FILE_URI and KINGFISHER_API_ITEM_URI to KINGFISHER_API_URI

odscjames:
This will make it easier to use new API end points in future commits without people having to add more to their settings

### DIFF
--- a/docs/use-scrapyd.rst
+++ b/docs/use-scrapyd.rst
@@ -80,7 +80,7 @@ Output - Kingfisher Process
 
 In order to use this, you must also set up the Disk output.
 
-In settings.py, make sure ``ITEM_PIPELINES`` includes ``KingfisherPostPipeline`` and that the 3 API variables are set to load from the environment. For example:
+In settings.py, make sure ``ITEM_PIPELINES`` includes ``KingfisherPostPipeline`` and that the 2 API variables are set to load from the environment. For example:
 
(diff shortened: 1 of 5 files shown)


odscjames:
Have added the new variable to Salt scripts for our servers.

odscjames:
I think so. 

> In the imaginary scenario that someone is using their own process endpoint, that's not going to work.

In that scenario, I'd expect them to write their own Pipeline component to push data. Because if they just try and reuse ours, they are already having to match the parameters we use, and keep up with any changes to that. Also, we are soon going to be using several different end points so they would have to match the purpose of those too. 

> And if we change the endpoint, it's going to need changing in the Python, not just in the environment.

Slight correction "if we change the endpoint, it's going to need changing in the Python INSTEAD OF the environment."

I see that as a good thing. If our API changes I want any changes to the sending code to be part of our code review process, and clearly in the code. And crucially, I want to make those changes ONCE. If we put these in new environmental variables and expect people to change them, then the changes have to happen in multiple places, with us having no insight as to whether it happened well or not. "Have you got the latest code?" is a much simpler question to ask people than "Have you got the latest code, oh and wait, which version were you on? You need to change env vars 4 and 8. Huh, not working. Oh, you didn't add env var 17! Of course."

PR #59, opened 2019-01-31
Send errors to new Error API

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/56


odscjames:
If so, this is a problem with the current code for the /api/v1/submit/file/ call - do you feel ok putting this p.r. in and I'll make an issue and try and look into that for both cases at the same time?

PR #96, opened 2019-02-12
New Moldova API

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/66

(It's called moldova_releases beacuse there is also a record end point we want to add, but we need to check some things first!)

### DIFF
--- a/kingfisher_scrapy/spiders/moldova_releases.py
+++ b/kingfisher_scrapy/spiders/moldova_releases.py
@@ -1,0 +1,20 @@
+import json
+import scrapy
+
+from kingfisher_scrapy.base_spider import BaseSpider
+
+
+class MoldovaReleases(BaseSpider):
+    name = 'moldova_releases'
+    start_urls = ['http://ocds.mepps.openprocurement.io/api/releases.json']
+
+    def parse(self, response):
+        json_data = json.loads(response.body_as_unicode())
+        if not self.is_sample():
+            if 'links' in json_data and 'next' in json_data['links']:
+                yield scrapy.Request(json_data['links']['next'])
+
+        yield {
+            'file_urls': [response.url],
+            'data_type': 'release_package'
+        }


kingfisher_scrapy/spiders/moldova_releases.py
```diff
@@ -0,0 +1,19 @@
...
+from kingfisher_scrapy.base_spider import BaseSpider
+
+
+class MoldovaReleases(BaseSpider):
+    name = 'moldova_releases'
+    start_urls = ['http://ocds.mepps.openprocurement.io/api/releases.json']
+
+    def parse(self, response):
+        json_data = json.loads(response.body_as_unicode())
+        if not self.is_sample():
```
BibianaC: Would this not be `if not (hasattr(self, 'sample') and self.sample == 'true'):`?

kingfisher_scrapy/spiders/moldova_releases.py
```diff
@@ -0,0 +1,19 @@
+import json
+import scrapy
+from kingfisher_scrapy.base_spider import BaseSpider
```
BibianaC: Could you order the imports, please?

kingfisher_scrapy/spiders/moldova_releases.py
odscjames: There is now a base method to avoid repeating code https://github.com/open-contracting/kingfisher-scrape/blob/master/kingfisher_scrapy/base_spider.py#L18

kingfisher_scrapy/spiders/moldova_releases.py
odscjames: What order are you expecting? I think that matches PEP8 https://www.python.org/dev/peps/pep-0008/#imports

kingfisher_scrapy/spiders/moldova_releases.py
BibianaC: Cool!

kingfisher_scrapy/spiders/moldova_releases.py
BibianaC: > You should put a blank line between each group of imports.

odscjames: Updated!

BibianaC: Thanks @odscjames !

BibianaC: Between json and scrapy it needs to be also a blank line. If you use PyCharm it will do it automatically for you in 'code/optimize imports'

PR #97, opened 2019-02-13
Change to correct license

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/28

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,29 @@
-MIT License
+BSD 3-Clause License
 
-Copyright (c) 2018 Yohanna Lisnichuk
+Copyright (c) 2018, Open Contracting Partnership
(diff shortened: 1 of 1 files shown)


odscjames:
@jpmckinney This is what has been used before so I think it's correct - can you confirm?

jpmckinney:
Yes - BSD-3 for Python code. I usually just put the first year, to avoid having to continuously change the end year.

odscjames:
Just had a quick Google (I thought putting the end year was recommended but apparently it's all a bit more complicated than that). I put it back to just start year then.

jpmckinney:
Updating the end year is generally only important for extending the copyright period, which isn't something we're interested in protecting (falling into public domain is fine).

PR #99, opened 2019-02-13
Filepath should have _sample at the end if it's a sample

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/52

### DIFF
--- a/kingfisher_scrapy/pipelines.py
+++ b/kingfisher_scrapy/pipelines.py
@@ -41,7 +41,10 @@ class KingfisherFilesPipeline(FilesPipeline):
         elif not media_ext or ('json' in content_type and media_ext != '.json'):
             media_ext = '.json'
         # Put files in a directory named after the scraper they came from, and the scraper starttime
-        return '%s/%s/%s%s' % (info.spider.name, start_time_str, media_guid, media_ext)
+        if hasattr(info.spider, 'sample') and info.spider.sample == 'true':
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/pipelines.py
```diff
@@ -41,7 +41,10 @@ def file_path(self, request, response=None, info=None):
         elif not media_ext or ('json' in content_type and media_ext != '.json'):
             media_ext = '.json'
         # Put files in a directory named after the scraper they came from, and the scraper starttime
-        return '%s/%s/%s%s' % (info.spider.name, start_time_str, media_guid, media_ext)
+        if hasattr(info.spider, 'sample') and info.spider.sample == 'true':
+            return '%s_sample/%s/%s%s' % (info.spider.name, start_time_str, media_guid, media_ext)
+        else:
```
BibianaC: the `else:` is not needed. If the `if` is false then it will move the next line.

odscjames: Oh yes, true. Tho personally I think it makes it easier to read - I find it clearer when skimming it

odscjames: I know that's very subjective :-)

jpmckinney: FYI, in most of our (and others') Python code, `else` clauses aren't used in such cases.

PR #100, opened 2019-02-13
After run finished, send finish to new end end point

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/57

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -1,6 +1,45 @@
 import scrapy
 import requests
 
+# This file contain base spiders all our spiders extend from, so we can add some custom functionality
+#
+# However, we need to do this for several base spiders. To avoid copying and pasting code between base spiders,
+#  some code is put into generic functions
+
+
+def _generic_get_start_time(called_on_class):
+    stats = called_on_class.crawler.stats.get_stats()
+    start_time = stats.get("start_time")
+    return start_time
+
+
+def _generic_is_sample(called_on_class):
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -1,6 +1,45 @@
...
+        response = requests.post(
+            called_on_class.crawler.settings['KINGFISHER_API_URI'] + '/api/v1/submit/end_collection_store/',
+            data=data,
+            headers=headers)
+
+        if not response.ok:
+            spider.logger.warning(
+                "Failed to post End Collection Store. API status code: {}".format(response.status_code))
+
+# Now we have our own base spiders (that use our generic functions)
```
BibianaC: I would delete this sentence.

kingfisher_scrapy/base_spider.py
```diff
@@ -1,6 +1,45 @@
 import scrapy
 import requests
 
+# This file contain base spiders all our spiders extend from, so we can add some custom functionality
+#
+# However, we need to do this for several base spiders. To avoid copying and pasting code between base spiders,
```
BibianaC: I would delete this sentence. The first line is enough to explain what is in the file.

odscjames:
If we delete those lines, there is nothing to explain why the generic functions are there. Someone could probably guess from looking at both classes, but I think it's good to comment to explain. And comments are usually good - we are trying to get more documentation at the moment!

BibianaC:
I'm all about comments but comments are good when they are needed. In this case is self explanatory.

PR #102, opened 2019-02-13
Old scrapers

odscjames:


odscjames:
Note I haven't changed .travis so flake8 runs on it, and the tests aren't copied. Given that we should not be developing this code, that seemed ok?

PR #107, opened 2019-02-21
Update colombia spider with more doc and cases

yolile:


odscjames:
if response.status == 503:
                wait_and_retry(response.url)

Should there not be an else after that? (Or something)

At the moment, if there is a 503 status, won't the function yield  both 

*  scrapy.Request(url)
* {               'file_urls': [response.url],             'data_type': 'release_package' }

when it should only yield the first?


ps. Is handle_httpstatus_list actually used?

yolile:
@odscjames you are rigth! I added the else. 
> ps. Is handle_httpstatus_list actually used?

So now, in their API when some service fail they return a malformed json, but they are going to update that and return a 503 error, so I'm just handling both cases

odscjames:
But is the variable handle_httpstatus_list actually used?

Should the if line use the variable, 

    if response.status in self.handle_httpstatus_list:

odscjames:
> that variable is a config for scrapy

Issue #91, opened 2019-02-07
Have a list of sources, and the state of each one somewhere


odscjames:
Maybe automatically generated, or just list in docs like the last one? (That seemed to work - the list doesn't change that often)

jpmckinney:
Should this be in a spreadsheet? I figure we shouldn't have to update the docs in order to update the state of a source.

odscjames:
I guess there are two types of changes we want to document:

  *  Changes to the external API, so something has changed even thought there is no code change. (Examples include a external API going off-line; that scraper is now dead even thought the code hasn't changed.)
  *  Changes to the code (Examples include adding the page parameter, as in #109)

A spreadsheet is good for the first point. A page in the docs like last time is good for the second point, as pull requests reviews can enforce that docs change as the code changes.

I don't know what is best - but it's good to have something, so I'm going to start with a page in the docs as that's what we had before, and that would help with the code we just merged. We can always move the information later, but it's good to just start gathering some info somewhere. I'll leave this ticket open so that discussion isn't forgotten.

jpmckinney:
Presumably closed by #110

jpmckinney:
https://kingfisher-scrape.readthedocs.io/en/latest/sources.html

PR #109, opened 2019-02-22
Add page argument in Colombia spider

yolile:
- Add a page argument to indicate in which page the spider should start 
- Change the error handling to ensure that the spider waits when an error occurs

### DIFF
--- a/kingfisher_scrapy/spiders/colombia.py
+++ b/kingfisher_scrapy/spiders/colombia.py
@@ -7,15 +7,18 @@ import time
 from kingfisher_scrapy.base_spider import BaseSpider
 
 
-def wait_and_retry(url):
-    time.sleep(120 * 60)
-    yield scrapy.Request(url)
-
-
 class Colombia(BaseSpider):
     name = 'colombia'
     start_urls = ['https://apiocds.colombiacompra.gov.co:8443/apiCCE2.0/rest/releases?page=1']
     handle_httpstatus_list = [503]
+    sleep = 120 * 60
+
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/colombia.py
```diff
@@ -7,15 +7,18 @@
...
     start_urls = ['https://apiocds.colombiacompra.gov.co:8443/apiCCE2.0/rest/releases?page=1']
     handle_httpstatus_list = [503]
+    sleep = 120 * 60
+
+    def start_requests(self):
+        base_url = 'https://apiocds.colombiacompra.gov.co:8443/apiCCE2.0/rest/releases?page=%d'
+        start_page = 1
+        if hasattr(self, 'page'):
+            start_page = int(self.page)
+        yield scrapy.Request(url=base_url % start_page, callback=self.parse)
```
odscjames: Just out of curiosity, is the callback in

    yield scrapy.Request(url=base_url % start_page, callback=self.parse)

really needed? I'm just surprised because the other yield statements don't have them!

odscjames:
Generally, it would be good if extra options are documented somewhere. I'll accept for now tho, and look at docs in #91

odscjames:
Just out of curiosity, is the callback in

```
yield scrapy.Request(url=base_url % start_page, callback=self.parse)
```

really needed? I'm just surprised because the other yield statements don't have them!

odscjames:
Ah, just noticed it's already been reviewed! Whoops :-)

PR #115, opened 2019-02-27
Rewrite to remove FilesPipeline

odscjames:
Deals with:

Can we stop Scrapy hitting each end point twice?
https://github.com/open-contracting/kingfisher-scrape/issues/114

Don't store local files named with a MD5
https://github.com/open-contracting/kingfisher-scrape/issues/45

Send better errors to Error API
https://github.com/open-contracting/kingfisher-scrape/issues/89

Also, this makes not relevant any more:

Do we only ever return one URL in file_urls?
https://github.com/open-contracting/kingfisher-scrape/issues/51

### DIFF
--- a/docs/use-scrapyd.rst
+++ b/docs/use-scrapyd.rst
@@ -61,13 +61,9 @@ Find out more in the `Scrapyd docs <https://scrapyd.readthedocs.io/en/latest/ove
 Output - Disk
 -------------
 
-In settings.py, make sure ``ITEM_PIPELINES`` includes ``KingfisherFilesPipeline`` and that ``FILES_STORE`` is set. For example:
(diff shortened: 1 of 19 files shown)


odscjames:
Also, this provides full logging in the database of non-200 HTTP responses.

Under the old way, it's possible a non-200 response would be dropped by scrapy, and not logged in our system.

yolile:
Can we have some kind of base parse method?

odscjames:
"Can we have some kind of base parse method?" I was thinking that, but I wasn't yet sure which bits might change per scraper and which bits might be the same and thus should be in a library. checked in chat - merging now and will bear in mind for future.

PR #130, opened 2019-03-18
Add indonesia scraper

### LINKED ISSUE #72: Port indonesia_bandung Scraper

yolile:
Ok, so that service returns a list of releases, no release_package and currently none of our models support that https://kingfisher-process.readthedocs.io/en/latest/data-model.html#data-types-for-files

And the old url https://birms.bandung.go.id/api/packages/year/2013 it is not working anymore either, closing and marking as blocked


odscjames:
Does this work with years?

2013 has 94 pages
2014 has 206 pages

But the way the for loop is, won't this just take one value and assume that value is the same for all years?

yolile:
So, I totally modified the spider acording to their documentation: https://birms.bandung.go.id/documentation

PR #159, opened 2019-05-21
Added Buenos Aires Spider

yolile:


kingfisher_scrapy/spiders/buenos_aires.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def start_requests(self):
+        yield scrapy.Request(
+            url='https://data.buenosaires.gob.ar/api/3/action/package_show?id=buenos-aires-compras',
+            meta={'type': 'meta'}
+        )
+
+    def parse(self, response):
+        if response.status == 200:
+            print(response.request.meta)
```
romifz: I would use `self.logger.info` here instead of `print`

yolile: I removed the print because it was unnecessary

PR #186, opened 2019-08-20
Port digiwhist Scrapers

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/74


odscjames:
Want to:
* Make sure whatever fix I did before for the bad line issue still works
* Try and write file info to disk such that local loading it later can be easier for people

odscjames:
> Make sure whatever fix I did before for the bad line issue still works

Thought so but just checked - this fix all happened on the process side so nothing to do here.

odscjames:
Felt the process_item func was getting a little long, so split up for readability reasons

odscjames:
Ready for review!

PR #197, opened 2019-09-25
add proxy option to scrapers

kindly:
Add a proxy option to kingfisher scrape.

### DIFF
--- a/docs/use-hosted.rst
+++ b/docs/use-hosted.rst
@@ -37,6 +37,13 @@ To start a run using 'sample mode', to obtain a small amount of data quickly:
 
 Update the note with your name, and anything else of interest.
 
+If downloading the data is a problem from the hosted server you can use a proxy server:
+
+.. code-block:: bash
+
+    $ curl http://localhost:6800/schedule.json -d project=kingfisher -d spider=canada_buyandsell -d proxy=true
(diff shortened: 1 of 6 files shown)


odscjames:
Haven't tested but on reading it looks good. 

My only comment is that the Hosted docs will say "You can do this!" and you won't be able to because the other half of the system (the actual Proxy and the KINGFISHER_PROXY_URL setting) won't have been done.

Could we put that in the docs hosted page but commented out for now? Then it's ready to go later. But I think the immediate user need has been cleared, so it may be a while before the other bit of work is prioritised.

Also that could go in the docs for the other pages - Standalone and Scrapyd - right now.

kindly:
@odscjames 

The proxy and the the KINGFISHER_PROXY_URL have beed deployed and are ready.  I thought I would do that first for the above reason.

So as soon as this pull request is merged and deployed the docs will work.

I have also just now added docs for the standalone a scrapyd parts.

odscjames:
I am minded to close this as having been superseded by #235 

It does do things in a different way, and if we use Proxies more in the future we may want to come back to this and work out which way is best. But for now, am closing.

jpmckinney:
I've removed the configuration that was added to deploy in September, to avoid future confusion.

PR #210, opened 2019-11-03
fix broken formatting for sources list

duncandewhurst:


docs/sources.rst
```diff
@@ -14,7 +14,8 @@ These sources are available as Scrapy spiders:
   *  chile_compra_releases
   *  chile_compra_records - you can pass the additional `year` parameter to this scraper, to indicate what year download.
   *  colombia - you can pass the additional `page` parameter to this scraper, to indicate what page to start on,
-  and also you can pass the additional `year` parameter to this scraper, to indicate what year download.
+  and also you can pass the additional `year` parameter to this scraper, to indicate what year download
```
jpmckinney: I think it's better to just put all the content on one line, instead of word wrapping.

duncandewhurst: Ah, I hadn't spotted it, that explains why the formatting was broken!

PR #212, opened 2019-11-11
store scrapy stats separately to full log due to potentially large fi…

odscjames:
…le size

https://github.com/open-contracting/kingfisher-scrape/issues/199


odscjames:
[Scrape: store scrapy stats separately to full log due to potentially large file size](https://trello.com/c/xEHgFO6O/114-scrape-store-scrapy-stats-separately-to-full-log-due-to-potentially-large-file-size)

PR #214, opened 2019-11-12
Convert Scrapy Spider to new pipeline: canada_montreal

odscjames:
https://github.com/open-contracting/kingfisher-scrape/issues/118

### DIFF
--- a/kingfisher_scrapy/spiders/canada_montreal.py
+++ b/kingfisher_scrapy/spiders/canada_montreal.py
@@ -6,7 +6,6 @@ from kingfisher_scrapy.base_spider import BaseSpider
 
 class CanadaMontreal(BaseSpider):
     name = 'canada_montreal'
-    start_urls = ['https://ville.montreal.qc.ca/vuesurlescontrats/api/releases.json']
     custom_settings = {
         'ITEM_PIPELINES': {
             'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
--- a/kingfisher_scrapy/spiders/canada_montreal.py
+++ b/kingfisher_scrapy/spiders/canada_montreal.py
@@ -1,32 +1,54 @@
 import json
+import scrapy
 
 from kingfisher_scrapy.base_spider import BaseSpider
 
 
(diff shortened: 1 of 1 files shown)


odscjames:
[Kingfisher Canada Scraper](https://trello.com/c/uQ3hKa0U/142-kingfisher-canada-scraper)

kingfisher_scrapy/spiders/canada_montreal.py
```diff
@@ -1,32 +1,54 @@
...
 
 from kingfisher_scrapy.base_spider import BaseSpider
 
 
-# This Spider uses the old system of pipelines! DO NOT USE IT AS AN EXAMPLE OF WHAT TO DO IN FUTURE SPIDERS!
-# Thank you.
 class CanadaMontreal(BaseSpider):
     name = 'canada_montreal'
-    start_urls = ['https://ville.montreal.qc.ca/vuesurlescontrats/api/releases.json?limit=1']
+    start_urls = ['https://ville.montreal.qc.ca/vuesurlescontrats/api/releases.json']
```
rhiaro: This is ignored when start_requests is overridden, so we can take this line out

kingfisher_scrapy/spiders/canada_montreal.py
```diff
@@ -1,32 +1,54 @@
...
-        }
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
     }
+    page_limit = 10000
+
+    def start_requests(self):
+        yield scrapy.Request(
+            url='https://ville.montreal.qc.ca/vuesurlescontrats/api/releases.json?limit=%d' % self.page_limit,
```
rhiaro: although since the URL is in more than one place, maybe keep it in start_urls and get it from there instead of repeating it?

odscjames: it's only 2 places, and also it's used in different ways, so i suspect any change would change how it's used too. I'll just do above point.

PR #226, opened 2019-12-02
Add Honduras API spider

aguilerapy:


kingfisher_scrapy/spiders/honduras_portal_records.py
```diff
@@ -0,0 +1,40 @@
...
+            yield scrapy.Request(file_url)
+
+            file_name = response.request.url
+
+            yield self.save_data_to_disk(json.dumps(json_data['results']).encode(), "page_" + file_name[53:] + ".json", data_type='record_list', url=response.request.url)
+
+        else:
+            yield {
+                'success': False,
+                'file_name': response.request.meta['releases.json'],
```
romifz: Please use an unique name here.

kingfisher_scrapy/spiders/honduras_portal_records.py
```diff
@@ -0,0 +1,40 @@
...
+        if response.status == 200:
+
+            json_data = json.loads(response.body_as_unicode())
+            file_url = json_data['next']
+
+            yield scrapy.Request(file_url)
+
+            file_name = response.request.url
+
+            yield self.save_data_to_disk(json.dumps(json_data['results']).encode(), "page_" + file_name[53:] + ".json", data_type='record_list', url=response.request.url)
```
romifz: Please try to use regular expressions to extract the page number only, to avoid unnecesary large file names.

kingfisher_scrapy/spiders/honduras_portal_releases.py
```diff
@@ -0,0 +1,44 @@
...
+
+            json_data = json.loads(response.body_as_unicode())
+            yield self.save_data_to_disk(
+                json.dumps(json_data['results']).encode(),
+                response.request.meta['kf_filename'],
+                data_type='release_list',
+                url=response.request.url
+            )
+
+            url = json_data['next']
```
yolile: @aguilerapy you should check if next is not null, in the last page it is: http://200.13.162.86/api/v1/release/?format=json&page=107995

kingfisher_scrapy/spiders/honduras_portal_records.py
```diff
@@ -0,0 +1,44 @@
...
+
+            json_data = json.loads(response.body_as_unicode())
+            yield self.save_data_to_disk(
+                json.dumps(json_data['results']).encode(),
+                response.request.meta['kf_filename'],
+                data_type='record_list',
+                url=response.request.url
+            )
+
+            url = json_data['next']
```
yolile: @aguilerapy you should check if next is not null, in the last page it is: http://200.13.162.86/api/v1/record/?format=json&page=94328

kingfisher_scrapy/spiders/honduras_portal_records.py
```diff
@@ -5,6 +5,8 @@
 
 class HondurasPortalRecords(BaseSpider):
     name = 'honduras_portal_records'
+    download_delay = 1.5
+    final = '94329'
```
romifz: The last page should calculated! Same for the releases spider.

kingfisher_scrapy/spiders/honduras_portal_records.py
yolile: @aguilerapy you can do that with something like:

```python
if 'next' in json_data and json_data['next']:

```

kingfisher_scrapy/spiders/honduras_portal_records.py
yolile: You can avoid this by checking if next is not null.

kingfisher_scrapy/spiders/honduras_portal_records.py
jpmckinney: Or shorter: `json_data.get('next')`

kingfisher_scrapy/spiders/honduras_portal_records.py
aguilerapy: Yes, I have a sentence that check if next is not null, this variable final only works with sample, I just define final previously

kingfisher_scrapy/spiders/honduras_portal_records.py
```diff
@@ -0,0 +1,48 @@
...
+                url=response.request.url
+            )
+
+            url = json_data['next']
+            if not url or self.is_sample():
+                return
+            else:
+                yield scrapy.Request(
+                    url,
+                    meta={'kf_filename': 'page%s.json' % url[53:]}
```
yolile: I would prefer to use `hashlib.md5(url.encode('utf-8')).hexdigest()` like the other scrapers instead of hardcoding the page position, but we can change that when the url change.

Issue #218, opened 2019-11-28
Write scrapers for new Paraguay API


duncandewhurst:
This is a high priority as OCP need to check the API is working so they can sign-off the project.

The documentation is here: [http://beta.dncp.gov.py/datos/api/v3/doc/](http://beta.dncp.gov.py/datos/api/v3/doc/)

It looks to use the same token/authentication model as the existing Paraguay API.

To get all the data, I think we need to write scrapers for only the following endpoints:

* /ocds/record/
* /ocds/releases/ocid

If I understood correctly, the list of ocids to pass as parameters can be obtained from the csv files here: [https://www.contrataciones.gov.py/datos/convocatorias](https://www.contrataciones.gov.py/datos/convocatorias)

@yolile - please confirm I got all the above correct, thank you!

duncandewhurst:
![](https://github.trello.services/images/mini-trello-icon.png) [Write scrapers for new Paraguay API](https://trello.com/c/QfFio5Jf/157-write-scrapers-for-new-paraguay-api)

yolile:
You can use the service
http://beta.dncp.gov.py/datos/api/v3/doc/search/processes
To list all the existing processes (only 2018 and 2019 are loaded for now)
and use the field compiledRelease.ocid to get fue full releases and records
with the two services that you already mentioned.
Also the token now expires after 15 minutes always, no with a certain
number of requests.

duncandewhurst:
@yolile suggested that @romifz can write the scraper

jpmckinney:
Noting that there are some issues mentioned in CRM 4864, though the scraper was updated in #227.

odscjames:
Finding: The Proxy was working on IPv4 only - I added IPv6 support. It now works for both HTTP and HTTPS.

> 
> root@process1 ~ # curl -x http://staging.docs.opencontracting.uk0.bigv.io:8888  http://example.com
> <!doctype html>
> <html>
> <head>
>     <title>Example Domain</title>
> .............
> 


> root@process1 ~ # curl -x http://staging.docs.opencontracting.uk0.bigv.io:8888  https://www.google.com
[...]

PR #227, opened 2019-12-05
Update Paraguay DNCP scrapers for new API

romifz:
#218

### DIFF
--- a/kingfisher_scrapy/middlewares.py
+++ b/kingfisher_scrapy/middlewares.py
@@ -11,19 +11,17 @@ class ParaguayAuthMiddleware(object):
     Both DNCP (procurement authority) and Hacienda (finance ministry) use an
     authentication protocol based on OAuth 2.0.
 
-    This middleware is an attempt to manage the protocol, which in short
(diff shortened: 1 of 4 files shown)


odscjames:
Did you know you can pass meta information with a request? https://github.com/open-contracting/kingfisher-scrape/blob/edf696807f5fe9027ec1cee50099d56ce92f9712/kingfisher_scrapy/spiders/moldova.py#L50 See use of the 'page' variable there. This might be able to replace the regular expression to get the page number from the URL.

odscjames:
But that is just a note for future work, feel free to merge now if you want.

romifz:
Thank you James! I noticed a mistake in the regular expression, it's fixed now. BTW I knew about the meta attribute in requests, I just like regular expressions 😅.

I just deleted a single character in a file, so I'll ask @yolile for approval to merge.

jpmckinney:
@romifz Can we [`urllib.parse`](https://docs.python.org/3/library/urllib.parse.html) the page, instead of using regular expressions? Regexes should only be used if there isn't a better way to parse a string.

Issue #82, opened 2018-08-17
Add Nepal scraper


robredpath:
The link we currently have doesn't work - we should check with Karolis re: current publication status

PR #239, opened 2019-12-23
Add Nepal scraper

aguilerapy:
#82

### DIFF
--- a/kingfisher_scrapy/spiders/nepal_portal.py
+++ b/kingfisher_scrapy/spiders/nepal_portal.py
@@ -1,0 +1,51 @@
+import hashlib
+import json
+import datetime
+
+import scrapy
+
+from kingfisher_scrapy.base_spider import BaseSpider
+
+
+class NepalPortal(BaseSpider):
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/nepal_portal.py
```diff
@@ -0,0 +1,51 @@
...
+        'ITEM_PIPELINES': {
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        current_year = 2013
+        if self.is_sample():
+            end_year = 2013
```
yolile: Currently  2013 has [no data](http://ppip.gov.np/bulk-download/2013), so maybe we can use 2018 as sample, i.e.:
```
end_year = 2018
current_year = 2018
```

aguilerapy: Okay!

Issue #230, opened 2019-12-09
Add Kenya - Makueni County scraper


duncandewhurst:
https://opencontracting.makueni.go.ke/api/ocds/package/all

PR #243, opened 2020-01-13
Add Kenya - Makueni scraper

aguilerapy:
#230

### DIFF
--- a/kingfisher_scrapy/spiders/kenya_makueni.py
+++ b/kingfisher_scrapy/spiders/kenya_makueni.py
@@ -1,12 +1,13 @@
 import hashlib
 import json
+import requests
 import scrapy
+
 from kingfisher_scrapy.base_spider import BaseSpider
 
 
 class KenyaMakueni(BaseSpider):
     name = 'kenya_makueni'
-    download_delay = 0.9
     custom_settings = {
         'ITEM_PIPELINES': {
             'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
@@ -16,27 +17,33 @@ class KenyaMakueni(BaseSpider):
 
     def start_requests(self):
         if self.is_sample():
-            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/ocid/ocds-muq5cl-18757'
+            page_size = 10
         else:
-            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all'
+            count = requests.get('https://opencontracting.makueni.go.ke/api/ocds/release/count')
+            page_size = int(count.text)
 
-        yield scrapy.Request(
-            url,
-            meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
-        )
+        if page_size > 300:
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageNumber={}'
+            for page in range(page_size):
+                yield scrapy.Request(
+                    url.format(page),
+                    meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
+                )
+
+        else:
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={}'
+            yield scrapy.Request(
+                url.format(page_size),
+                meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
+            )
 
     def parse(self, response):
         if response.status == 200:
-            if self.is_sample():
-                data_type = 'release_package',
-            else:
-                data_type = 'release_package_list',
-
             json_data = json.loads(response.body_as_unicode())
             yield self.save_data_to_disk(
                 json.dumps(json_data).encode(),
                 response.request.meta['kf_filename'],
-                data_type=data_type,
+                data_type='release_package_list',
                 url=response.request.url
             )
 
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,49 @@
...
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        if self.is_sample():
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/ocid/ocds-muq5cl-18757'
+        else:
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all'
```
yolile: I noticed that that service has the 'pageNumber' and 'pageSize' query params, that are set to 0 and 300 by default. The service https://opencontracting.makueni.go.ke/api/ocds/release/count says that there are 160 releases, so calling **now** https://opencontracting.makueni.go.ke/api/ocds/package/all without parameters indeed gives you all the releases packages. BUT if they, in a future, add more releases, this scraper will give us just the first 300 ones and not all of them. Maybe you can call the service https://opencontracting.makueni.go.ke/api/ocds/release/count first to know how many releases are and then call the other iterating the pageNumber as appropriate

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,49 @@
...
+    custom_settings = {
+        'ITEM_PIPELINES': {
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        if self.is_sample():
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/ocid/ocds-muq5cl-18757'
```
yolile: Maybe we can get the first 10 releases instead of a fixed ocid with https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize=10

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,49 @@
...
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all'
+
+        yield scrapy.Request(
+            url,
+            meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
+        )
+
+    def parse(self, response):
+        if response.status == 200:
+            if self.is_sample():
```
yolile: if we change the sample to https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize=0 this will no longer be needed

kingfisher_scrapy/spiders/kenya_makueni.py
aguilerapy: Apparently the pageNumber parameter doesn't work, but add it anyway

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,56 @@
...
+    def start_requests(self):
+        if self.is_sample():
+            page_size = 10
+        else:
+            count = requests.get('https://opencontracting.makueni.go.ke/api/ocds/release/count')
+            page_size = int(count.text)
+
+        if page_size > 300:
+            url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageNumber={}'
+            for page in range(page_size):
```
yolile: 'https://opencontracting.makueni.go.ke/api/ocds/release/count' gives you the total number of available releases, then, if each page has 300 releases (pageSize=300) you will have = **total/page_size** pages, so the range should be `(ceil(total/page_size))`.

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,56 @@
...
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        if self.is_sample():
+            page_size = 10
+        else:
+            count = requests.get('https://opencontracting.makueni.go.ke/api/ocds/release/count')
+            page_size = int(count.text)
```
yolile: this should be `total=int(count.text)`

kingfisher_scrapy/spiders/kenya_makueni.py
yolile: If I use pageNumber with a smaller pageSize (https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize=10&pageNumber=2) it works. The default pageNumber is 300 and there are only 160 releases now, so for that if you you a pageSize bigger than 160 you will only get 1 page.

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def start_requests(self):
+        if self.is_sample():
+            total = 10
+        else:
+            count = requests.get('https://opencontracting.makueni.go.ke/api/ocds/release/count')
+            total = int(count.text)
+
+        url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={}&pageNumber={}'
+        page_size = 10
```
yolile: @aguilerapy 10 was just an example, you can keep the page_size at 300 :+1:

PR #244, opened 2020-01-13
Add quien es quien scraper

aguilerapy:
I could not find a simple way to get all the data from de [API](https://quienesquienapi.readthedocs.io/en/latest/).

I believe that `count` in `data/collections/contracts` [here](https://api.quienesquien.wiki/v2/sources) its the total amount of contracts and therefore I use it to iterating the `offset` parameter and get all the contracts.

### DIFF
--- a/kingfisher_scrapy/spiders/quien_es_quien.py
+++ b/kingfisher_scrapy/spiders/quien_es_quien.py
@@ -1,15 +1,15 @@
 import hashlib
 import json
-import math
 import requests
-
 import scrapy
+from math import ceil
 
 from kingfisher_scrapy.base_spider import BaseSpider
 
 
-class QuienEsQuienWiki(BaseSpider):
-    name = 'quien_es_quien_wiki'
+class QuienEsQuien(BaseSpider):
+    name = 'quien_es_quien'
+    download_delay = 0.9
     custom_settings = {
         'ITEM_PIPELINES': {
             'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/quien_es_quien.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def start_requests(self):
+        if self.is_sample():
+            url = 'https://api.quienesquien.wiki/v2/contracts?limit=10'
+            yield scrapy.Request(
+                url,
+                meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
+            )
+        else:
+            url = 'https://api.quienesquien.wiki/v2/contracts?limit=1000&offset={}'
```
yolile: maybe you can put the fixed "1000" limit as a constant

kingfisher_scrapy/spiders/quien_es_quien.py
```diff
@@ -0,0 +1,55 @@
...
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        if self.is_sample():
+            url = 'https://api.quienesquien.wiki/v2/contracts?limit=10'
+            yield scrapy.Request(
+                url,
+                meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'}
```
yolile: for samples you can use the name "sample.json"

PR #245, opened 2020-01-14
Add Portugal scraper

aguilerapy:
The zip file available at https://dados.gov.pt/pt/datasets/ocds-portal-base-www-base-gov-pt/ has multiple JSON root elements, like this: {record package}{record package}{record package}...

Assuming the file type: record_package_json_lines.

### DIFF
--- a/kingfisher_scrapy/spiders/portugal.py
+++ b/kingfisher_scrapy/spiders/portugal.py
@@ -1,5 +1,4 @@
 import hashlib
-import json
 from io import BytesIO
 from zipfile import ZipFile
 
--- a/kingfisher_scrapy/spiders/portugal.py
+++ b/kingfisher_scrapy/spiders/portugal.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 from io import BytesIO
 from zipfile import ZipFile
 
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/portugal.py
```diff
@@ -0,0 +1,44 @@
...
+    download_warnsize = 0
+    custom_settings = {
+        'ITEM_PIPELINES': {
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        url = 'https://dados.gov.pt/pt/datasets/r/ae76989b-d413-4f00-b1f0-0de1782212b6'
```
yolile: Instead of having a fixed URL that will probably change (when they upload a new file), you can do something like what was done for mexico_inai and buenos_aires scrapers where they also have an open data portal to scrape. From https://dados.gov.pt/pt/apidoc/#!/datasets/list_datasets, You can query the url:
https://dados.gov.pt/api/1/datasets/?q=ocds&organization=5ae97fa2c8d8c915d5faa3bf&page_size=20
where 5ae97fa2c8d8c915d5faa3bf is the publisher organization id and the query is "ocds". 
Then you can go to data.resources-> and search for resources's urls that has 'OCDS' in their description.

Issue #121, opened 2019-02-28
Add Uganda scraper


odscjames:
~~The new pipeline has several benefits we want; better filenames, error logs, not hitting API end points twice sometimes ...~~

odscjames:
On hold, see CRM 5332

jpmckinney:
I think the CRM issue indicates that we can try new API endpoint (the fact that it 500s sometimes shouldn't be a blocker, as other APIs are also unstable).

I don't see much immediate similarity between the APIs. Since it's the last "old pipeline" scraper, I'm removing it, so that we can remove the old pipeline code.

For reference, the last version of the old scraper is https://github.com/open-contracting/kingfisher-scrape/blob/fffbde8/kingfisher_scrapy/spiders/uganda.py

jpmckinney:
See issue about old scraper if relevant: #85

PR #248, opened 2020-01-16
Add Uganda scraper

aguilerapy:
#121

### DIFF
--- a/kingfisher_scrapy/spiders/uganda_releases.py
+++ b/kingfisher_scrapy/spiders/uganda_releases.py
@@ -17,8 +17,9 @@ class Uganda(BaseSpider):
     }
 
     def start_requests(self):
-        url = 'https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/planning?fy={}&pde={}'
+        url = 'https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/{}?fy={}&pde={}'
         url_pdes = 'https://gpp.ppda.go.ug/adminapi/public/api/pdes?page={}'
+        tags = ['planning', 'tender', 'award', 'contract']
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -0,0 +1,74 @@
...
+    download_delay = 0.9
+    custom_settings = {
+        'ITEM_PIPELINES': {
+            'kingfisher_scrapy.pipelines.KingfisherPostPipeline': 400
+        },
+        'HTTPERROR_ALLOW_ALL': True,
+    }
+
+    def start_requests(self):
+        url = 'https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/planning?fy={}&pde={}'
```
yolile: @aguilerapy it seems that, like in the [old api](https://github.com/open-contracting/kingfisher-scrape/blob/fffbde8/kingfisher_scrapy/spiders/uganda.py) they also have tender, award and contract releases.

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -0,0 +1,74 @@
...
+        url_pdes = 'https://gpp.ppda.go.ug/adminapi/public/api/pdes?page={}'
+        pdes_fdy_checks = []
+
+        if self.is_sample():
+            total_pages = 1
+        else:
+            pages = requests.get('https://gpp.ppda.go.ug/adminapi/public/api/pdes')
+            total_pages = pages.json()['data']['last_page']
+
+        for page_number in range(1, total_pages + 1):
```
yolile: @aguilerapy maybe you can simplify this a little bit using iterators instead of looping by the index

PR #260, opened 2020-01-30
Refactor pipelines

jpmckinney:
Like with #259, I added full test coverage, then refactored pipelines.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ install:
 script:
   - flake8
   - isort
-  - python -m pytest --cov kingfisher_scrapy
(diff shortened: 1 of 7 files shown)


jpmckinney:
As with #259, this PR makes no changes in behavior. It makes no changes to class APIs besides:

1. renaming `api_local_directory` to `directory`, though I'm not sure that this property was intended to be part of `KingfisherPostPipeline`'s public API anyhow.
2. promoting `BaseSpider.get_start_time()` to a public API method.

It also adds a small Kingfisher Process API client, to centralize that logic in one place.

PR #264, opened 2020-01-30
Always DropItem, even if the Kingfisher Process API responds with an error code

### LINKED ISSUE #88: Should we always be raising DropItem at the end of the pipeline?
From #59 - @rhiaro says:

> I think you need to raise `DropItem` regardless of whether the request was successful, otherwise it's going to pass the API response through the pipeline and throw up some more confusing errors

jpmckinney:
closes #88

### DIFF
--- a/kingfisher_scrapy/pipelines.py
+++ b/kingfisher_scrapy/pipelines.py
@@ -72,8 +72,9 @@ class KingfisherPostPipeline:
 
     def _post(self, item, spider, method, *args, name='API'):
         response = getattr(spider.client, method)(*args)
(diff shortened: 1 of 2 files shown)


odscjames:

jpmckinney:
Before merging, we might want to discuss in #88 whether it's better to just connect an item_scraped signal, and not have a pipeline at all.

jpmckinney:
I explored the direction in #88 a bit, and I'm convinced a signal in an extension is the better way to go, and that we're not using the pipeline for its usual purpose.

PR #265, opened 2020-01-30
Raise an error if file_name is missing

### LINKED ISSUE #237: Consider raising an error if 'file_name' not yielded or is None
This only occurs in `chile_base`.

See discussion in https://github.com/open-contracting/kingfisher-process/issues/47

I think the place to raise an error is `process_item`.

jpmckinney:
…as Kingfisher Process uses it as part of a unique key.

- Add a base class for all of this project's exceptions

closes #237

### DIFF
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -1,2 +1,10 @@
-class AuthenticationFailureException(Exception):
-    pass
+class KingfisherScrapyError(Exception):
+    """Base class for exceptions from within this application"""
+
+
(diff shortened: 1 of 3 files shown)


jpmckinney:
Although I wrote a test, I'm not 100% sure what the behavior is when running the spider in a real environment. For instance, I don't know if we need to connect an item_error signal: https://docs.scrapy.org/en/latest/topics/signals.html#item-error

odscjames:

jpmckinney:
Closing as this will require a new PR after #88 is resolved.

Issue #266, opened 2020-01-30
digiwhist_poland: Check that download_maxsize doesn't need override


jpmckinney:
By default, the max is 1GB: https://docs.scrapy.org/en/latest/topics/settings.html#download-maxsize

According to https://opentender.eu/download, Poland is 2.4 GB.

PR #273, opened 2020-01-31
Add download_maxsize in digiwhist_poland

aguilerapy:
#266


jpmckinney:
@aguilerapy Out of curiosity, does the spider fail without this change?

aguilerapy:
@jpmckinney yes it does

PR #271, opened 2020-01-31
Make spider arguments easier to access

jpmckinney:
Easier to review one commit at a time, to not be distracted by changes to specific spiders.

PR #274, opened 2020-01-31
Use item scraped signal instead of pipeline

### LINKED ISSUE #88: Should we always be raising DropItem at the end of the pipeline?

jpmckinney:
Closes #88 

This includes #271 to avoid merge conflicts, and just adds one commit: https://github.com/open-contracting/kingfisher-scrape/pull/274/commits/efc96b0250ede794926f4767db3d9b851ca96a14 The diff doesn't show pipelines.py and extensions.py side-by-side, but they are largely the same. As you can tell from the test files (which are shown side-by-side), little has changed.

The only change in behavior is that there is a different error message if the Kingfisher Process API URL and/or key aren't configured.

This also moves the part of the spider_closed signal that interacts with the API from `base_spider.py` into this extension. That way, all interaction with the API is here.

It also adds a base class from which all of this project's exceptions can inherit.


jpmckinney:

Issue #275, opened 2020-01-31
Do we need kingfisher.collectioninfo and kingfisher-finished.collectioninfo?


jpmckinney:
If we do `self.crawler.stats.set_value('kingfisher/source', self.name)`, etc. then this information will instead appear in the `*_report.log` files, which I think is fine, and takes advantage of parts of the Scrapy framework, rather than inventing our own solution.

jpmckinney:
@yolile @romifz @duncandewhurst @pindec @mrshll1001 Please contribute your thoughts, including whether you've ever used the kingfisher.collectioninfo or kingfisher-finished.collectioninfo files that are written for each crawl.

yolile:
I didn't even know that those files exist, so I never used them

duncandewhurst:
I don't use them!

romifz:
I neved used then neither.

odscjames:
I put these and the *.fileinfo files in.

The purpose was so we could write a better local-load/restore system at some point.

Currently, if you have a bunch of files on disk from a year ago and you want to load them back into the system you can, but you have to manually work out and pass various attributes. You have to pass the type of the files or it totally breaks.  Some information, like the original URL of the file is lost and can't be restored. 

I quickly threw these files in when I was changing something else with the system with the idea that at some point we could write a much better restore function; you'd simply point it at a directory, and it would be able to restore everything else including various bits of meta information.

Its true some of this information (all of it, possibly?) should appear in the log of the scrape, but those logs aren't currently stored with the files so they could easily be lost.  Also getting it to dump extra info alongside makes for much easier reading later than writing something to parse a log reliably and error free.

jpmckinney:
This issue is just about the collectioninfo files, not the fileinfo files (which can be discussed separately in #307).

As noted in the PR #290:

> The source, data version, and sample status are already represented in the file path (`<source>[_sample]/<data_version>`). The fact that the spider finished (previously represented by the existence of `kingfisher-finished.collectioninfo`) is already represented in the crawl log.

> After this PR, the note and end time won't be represented anywhere on the filesystem, but, from Kingfisher Scrape's perspective, I don't think we care that much about notes (most are just who started the scrape) or the end time (#275 suggested they could just be recorded as crawl stats). 

> The note and end time are still sent to Kingfisher Process, where we do care about them.

[...]

Issue #283, opened 2020-02-04
Make it easy to determine the latest release date across all sources


jpmckinney:
We don't want to have to download all the data from all the sources to determine this. Some options:

* If possible, spiders should be written so that the most recent data is collected first. That way, we can use [CLOSESPIDER_ITEMCOUNT](https://docs.scrapy.org/en/latest/topics/extensions.html#std:setting-CLOSESPIDER_ITEMCOUNT) like in #84 to just download one item.
* If that's not possible for some sources, we'll have to explore alternatives (perhaps specific to that spider), or just exclude that spider from consideration (and instead follow a manual process to determine the latest release date).

Ideally, a new command could [run all the spiders simultaneously](https://docs.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process). The spiders would be configured (perhaps with a spider argument) to just collect one item (with the latest release date). The item pipeline would transform the item to just return the source_id and release date (if the spider argument isn't set, this item pipeline would just pass-through the item). We would need to edit our extensions #274 #277 to do nothing with these special items (e.g. by testing whether the items have the fields needed by those extensions). The new command could be run locally.

The end result would be that OCP can run this command and easily determine which publishers have published any data within the calendar year, which we need for MEL.

Issue #277, opened 2020-01-31
Use item_scraped signal in extension, instead of writing files in spider


jpmckinney:
*Original issue title: Explore Item Exporters instead of bespoke system for writing items*

*Original issue description: I haven't had time to look at this closely, but on the surface it seems to do the same thing. Ideally, we can substitute Item Exporters without causing any change in behavior.*

https://docs.scrapy.org/en/latest/topics/exporters.html

jpmckinney:
Item exporters are designed to open a file, [start_exporting()](https://docs.scrapy.org/en/latest/topics/exporters.html#scrapy.exporters.BaseItemExporter.start_exporting), [export_item()](https://docs.scrapy.org/en/latest/topics/exporters.html#scrapy.exporters.BaseItemExporter.export_item) however many times, then [finish_exporting()](https://docs.scrapy.org/en/latest/topics/exporters.html#scrapy.exporters.BaseItemExporter.finish_exporting). In our case, we want one file per item. Item exporters aren't designed for that; it would be a huge overhead to have one item exporter per item.

As for files pipelines, that was used in the 'old pipeline' #115 #51. File pipelines are designed for a case where a spider gets a resource, and wants to schedule the download of that resource's media (images, PDFs, etc.); by design, a resource might lead to further requests (e.g. a next page link in the resource), whereas media don't lead to further requests. However, in our case, the resource *is also* the media that we want to download and store (most of the time). In the 'old pipeline', this led to duplicate requests (once for the resource by the spider, and once for the media by the files pipeline). While Scrapy can cache requests, in principle, we shouldn't use the files pipeline, because we don't have a clean separation between resources and media. (NB: We still use the `FILES_STORE` variable, which is meaningful to the files pipeline, but for different purposes.)

----

Taking a step back: Right now, the spider writes files itself, which doesn't make sense in [Scrapy's architecture](https://docs.scrapy.org/en/latest/topics/architecture.html#data-flow). The spider is only expected to yield items, which then go through an item pipeline to be dropped, errored, or passed through, and which then send an `item_scraped` signal.

We should reserve the item pipeline for the types of operations for which it was intended, e.g. deduplication, filtering, cleaning, validating, etc.

Whatever items make it through the pipeline can then be written to disk, by connecting the `item_scraped` signal to a new extension (which will be given higher priority in settings.py than the Kingfisher Process API extension #274).

yolile:
[...]

Issue #84, opened 2018-10-23
Add "has sample mode" option per source


odscjames:
This is needed for #207 - none of the Digiwhist Sources have sample modes.

When someone tries to run a Source in Sample mode that has no Sample, ignore this and show a nice message to them instead.

This is particularly important for when people run "run --all --sample"

(from discussion with @duncandewhurst )

jpmckinney:
I had a quick look, and this indeed still needs to be done, i.e.:

* each spider can have a class variable that indicates whether it supports sampling (if we add more CLI options in the future, we might want to use something other than a long series of variables)
* if a spider doesn't support sampling and `--sample` is provided, then error. Alternately, show a message saying sampling isn't supported, and don't set the collection's sample to true.

207 refers to https://github.com/open-contracting/kingfisher/issues/207 (issue had been moved between repositories)

odscjames:
We can maybe pause this at the moment tho - before this was important because someone might have run "run --all --sample" expecting a small download for each.

But with the new system there isn't currently a way to run all at once, and when we set one up we can make sure it works well in sample mode or non-sample mode. Depending on how the new run-all options works, that may alter the best technical way to achieve this.

jpmckinney:
I think we can perhaps have a global solution that uses `CLOSESPIDER_ITEMCOUNT` to close the spider once the sample size has been met.

odscjames:
> I think we can perhaps have a global solution that uses CLOSESPIDER_ITEMCOUNT to close the spider once the sample size has been met.

I would be wary of this because the number of items being small is only one thing of several I think people would expect on activating a sample mode. They would also expect that their machine would only download a reasonable amount of data. Under a global CLOSESPIDER_ITEMCOUNT scheme, someone trying to get a sample of France's Digiwhist data will still download 1GB.

jpmckinney:
digiwhist_base presently downloads the entire file, then breaks after a given number of lines. This issue isn't about making those per-spider improvements. I've opened a new issue about streaming a download so that we can gunzip only part of it: #312

jpmckinney:
Since Scrapy can in some cases send many requests before the CLOSESPIDER_ITEMCOUNT is reached, what we can do (either in addition to using CLOSESPIDER_ITEMCOUNT, or as an alternative) is add an item pipeline with a high priority that counts the number of items it has seen and then raises DropItem for any further items. (We might want to not count FileError items.)

[...]

PR #290, opened 2020-02-06
Remove collectioninfo files

jpmckinney:
Includes #274 to avoid a merge conflict.

#275 indicated that no one uses the `*.collectioninfo` files (or knew they existed).

The source, data version, and sample status are already represented in the file path (`<source>[_sample]/<data_version>`). The fact that the spider finished (previously represented by the existence of `kingfisher-finished.collectioninfo`) is already represented in the crawl log.

After this PR, the note and end time won't be represented anywhere on the filesystem, but, from Kingfisher Scrape's perspective, I don't think we care that much about notes (most are just who started the scrape) or the end time (#275 suggested they could just be recorded as crawl stats). 

The note and end time are still sent to Kingfisher Process, where we do care about them.


jpmckinney:

odscjames:
I commented on #275 with my purpose in adding these

odscjames:
Following the discussion in #275, I'd question whether this should be done yet.

#308 isn't done yet, and until it does we could lose some info if trying to restore, such as whether a scrape actually finished successfully or not. Maybe not much info, but given that it's only 2 small files on disk it doesn't seem worth rushing to remove them. The code is nicely in the base_spider.py alongside the code for writing the actual data to disk, so it's not too messy.

But I can review if you want to progress now - I think a rebase is needed first tho.

jpmckinney:
Let's merge it as the spider shouldn't be writing files to disk (continued in #277), since doing so circumvents the item pipeline entirely, which prevents us from taking advantage of Scrapy features that would be helpful in closing some other issues e.g. #283 #84.

#308 is easy to close, and I don't anticipate we'll delete log files or kill the server before we back them up.

yolile:
@jpmckinney the test coverage decreased

jpmckinney:
@yolile Yes, that's because we're deleting code that was tested, so the total coverage is dropping down closer to the average.

Issue #211, opened 2019-11-07
Add France scraper releases


pindec:
We need a scraper for France.

They have an alpha release of their OCDS here: https://www.data.gouv.fr/fr/datasets/r/68bd2001-3420-4d94-bc49-c90878df322c

data URL should be static - it's linked from this page https://www.data.gouv.fr/fr/datasets/fichiers-consolides-des-donnees-essentielles-de-la-commande-publique/ 

The data contains a single OCDS release package.

The intention is that this will be updated daily in future but is not yet being updated, so this is not an urgent request.

pindec:
![](https://github.trello.services/images/mini-trello-icon.png) [Scrape: write a scraper for France](https://trello.com/c/MhrhvFDn/139-scrape-write-a-scraper-for-france)

PR #311, opened 2020-02-17
Add France scraper

aguilerapy:
#211


kingfisher_scrapy/spiders/france.py
```diff
@@ -0,0 +1,62 @@
...
+
+from kingfisher_scrapy.base_spider import BaseSpider
+
+
+class France(BaseSpider):
+    name = "france"
+
+    def start_requests(self):
+        yield scrapy.Request(
+            url='https://www.data.gouv.fr/api/1/datasets/',
```
yolile: You can use the Etalab organization (the publisher) id to get the json url faster (in the first page) https://www.data.gouv.fr/api/1/datasets/?organization=534fff75a3a7292c64a77de4

kingfisher_scrapy/spiders/france.py
```diff
@@ -0,0 +1,62 @@
...
+                    description = resource.get('description')
+                    if description and (description.count("OCDS") or description.count("ocds")):
+                        url = resource.get('url')
+                        yield scrapy.Request(
+                            url,
+                            meta={'kf_filename': hashlib.md5(url.encode('utf-8')).hexdigest() + '.json'},
+                        )
+                        break
+                else:
+                    continue
```
yolile: you should add a sample mode even if now there are only one json file anyway

Issue #313, opened 2020-02-18
digiwhist_france: download_maxsize needs override


jpmckinney:
The file is presently 1005MB, so it's likely to exceed the 1024MB limit in the future.

We can do the same as in #266, but we might instead want to raise the limit globally in settings.py.

jpmckinney:
Closed by #314

Issue #276, opened 2020-01-31
Add integration tests


jpmckinney:
The current tests focus on individual components. We'd have greater confidence with a few, limited integration tests. We can perhaps test against quotes.toscrape.com like in the Scrapy tutorials, to check that our signals are connected, extensions are loaded, etc.

jpmckinney:
To test the Kingfisher Process API integration, we can send requests to httpbin (and use pytest-vcr).

jpmckinney:
Noting that we did have one instance where signals were not connected for almost 2 months: #659

jpmckinney:
Since we're using all extensions in scheduled operations, I think we can skip on additional integration tests.

PR #314, opened 2020-02-19
Raise download_maxsize

aguilerapy:
Raise the limit globally for download_maxsize to 4GB in settings.py #313

### DIFF
--- a/kingfisher_scrapy/settings.py
+++ b/kingfisher_scrapy/settings.py
@@ -29,6 +29,10 @@ CONCURRENT_REQUESTS = 32
 # See https://docs.scrapy.org/en/latest/topics/settings.html#download-delay
 # See also autothrottle settings and docs
 #DOWNLOAD_DELAY = 3
+
+# The maximum response size (in bytes) that downloader will download (default: 1073741824):
+DOWNLOAD_MAXSIZE = 4000000000
+
 # The download delay setting will honor only one of:
 CONCURRENT_REQUESTS_PER_DOMAIN = 2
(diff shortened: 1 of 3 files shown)


kingfisher_scrapy/settings.py
```diff
@@ -29,6 +29,10 @@
 # See https://docs.scrapy.org/en/latest/topics/settings.html#download-delay
 # See also autothrottle settings and docs
 #DOWNLOAD_DELAY = 3
+
+# The maximum response size (in bytes) that downloader will download (default: 1073741824):
+DOWNLOAD_MAXSIZE = 4000000000
```
yolile: @aguilerapy maybe you should remove the custom config in digiwhist_poland then

tests/test_settings.py
```diff
@@ -0,0 +1,7 @@
+import kingfisher_scrapy.settings
+
+
+class TestSettings():
+    name = 'test'
+    both_name = kingfisher_scrapy.settings.BOT_NAME
+    assert both_name == kingfisher_scrapy.settings.BOT_NAME
```
yolile: @jpmckinney should we exclude the settings file from travis? or do you think that there is a better way to test the settings?

jpmckinney: Why exclude it? Just to increase coverage? If so, I think we can leave it for now until #276 is resolved.

yolile: Mainly because this test as it is now doesnt test anything so, yes, lets keep this open until #276 is resolved

jpmckinney: Right - then let's remove this test file from the PR.

Issue #318, opened 2020-02-19
LinksSpider code review


jpmckinney:
Follow-up to #286 

* Instead of using multiple inheritance in the spiders, I think we should just have `LinksScraper` inherit from `BaseSpider`.
* Shouldn't we `yield from self.parse_next_link` instead of `return self.parse_next_link` ? (And, for moldova_records, instead of `yield self.parse_next_link`.)
* Instead of passing `self.sample` to `parse_next_link`, can't `parse_next_link` just use `self.sample` directly? Same for `self.save_response_to_disk`.

Great work on the recent PRs! For bigger changes to files outside the `spiders` directory (like #286), it might be good to request my review after @yolile's review, to go over comments like above.

yolile:
Sorry @jpmckinney my fault! I reviewed the code and noticed that we can reuse even more the code in the LinkSpider class but didnt want to change so much the @aguilerapy implementation so I didnt update also the inheritance etc. Done now!

PR #319, opened 2020-02-19
Update LinkSpider class and uses

yolile:
Refs #318


jpmckinney:
Looks good - I think there's just a flake8 issue.

PR #320, opened 2020-02-21
Fix indonesia scraper

yolile:


jpmckinney:
Instead of `.get()`, should we use `[]`? We should only ever use `.get()` if we expect the key to be missing.

Issue #323, opened 2020-02-26
paraguay_base/hacienda: Implementation questions / review


jpmckinney:
* Why is it necessary to call `self.crawler.engine.crawl`? Can we add a code-level comment?
* The 2nd `logger` line for "Requesting access token…" should be inside the `else` branch
* paraguay_hacienda includes the `clientSecret` value. Should this be kept secret, and provided as a spider argument instead?

yolile:
> Why is it necessary to call self.crawler.engine.crawl? Can we add a code-level comment?

it is necessary because that method is called througth a spider start_request or parse method, or other method where the request it is send to the scrapy request queue, so if we yield the request instead the request is never called.  With `self.crawler.engine.crawl` the request it is inserted in the queue and be called.

> paraguay_hacienda includes the clientSecret value. Should this be kept secret, and provided as a spider argument instead?

This is actually a enviroment variable already, but I noticed that it is also hardcoded in one line, we should change that

@aguilerapy please update paraguay_base and paraguay_hacienda with the suggested changes

jpmckinney:
I did a quick search for other Scrapy Oauth work, and there are a few middlewares mentioned in this thread: https://groups.google.com/forum/#!msg/scrapy-users/EUzuuy6oaeE/7tmZBsJLDAAJ

They use downloader middlewares instead, in order to intercept the requests. For example: https://github.com/scrapy/scrapy/compare/master...juanriaza:oauth-draft?expand=1

This might be a more correct use of Scrapy than calling `crawler.engine.crawl` directly.

aguilerapy:
@jpmckinney I think those solutions aren't for our case because the token for Paraguay can expire. Therefore, we use `crawler.engine.crawl` because we must stop and send an "new valid token" request for a new token when it expires, with `crawler.engine.crawl` the request is inserted into the queue and called immediately, so the scraper continues where it stops with the new token.

We use download middlewares to intercept the requests and add the token with `request.headers ['Authorization'] = spider.access_token`, similar to the other Scrapy Oauth work, but now I delete the call to `crawler.engine.crawl`.

What I did is modify our `process_request()` in ParaguayAuthMiddleware class:

1. Save the last request that has the expired token or doesn't have a token yet (the first time).
2. Return that "emergency" request in `process_request()` with `return spider.request_access_token()`
3. Continue scraping where it stops after obtaining the new token with `yield self.last_request` (calling the last saved request) in `parse_access_token()`

[...]

PR #324, opened 2020-02-27
Update paraguay scrapers

aguilerapy:
#323
- [x] fix middlewares
- [x] move logger
- [x] fix clientSecret value


kingfisher_scrapy/spiders/paraguay_hacienda.py
```diff
@@ -13,17 +13,18 @@ class ParaguayHacienda(BaseSpider):
...
     release_ids = []
-    request_limit = 10000
     request_time_limit = 14.0
 
     custom_settings = {
         'DOWNLOADER_MIDDLEWARES': {
            'kingfisher_scrapy.middlewares.ParaguayAuthMiddleware': 543,
         },
         'CONCURRENT_REQUESTS': 1,
+        'DUPEFILTER_DEBUG': True,
```
jpmckinney: Do we want this, or was it only here for testing?

kingfisher_scrapy/middlewares.py
```diff
@@ -66,16 +66,21 @@ def process_request(self, request, spider):
             spider.crawler.stop()
             raise scrapy.exceptions.IgnoreRequest()
         request.headers['Authorization'] = spider.access_token
-        if self._expires_soon(spider):
+        if self._expires_soon(spider) or not spider.access_token:
```
jpmckinney: Why not put the `not spider.access_token` logic in `_expires_soon`?

kingfisher_scrapy/middlewares.py
```diff
@@ -66,16 +66,21 @@ def process_request(self, request, spider):
...
+            # SAVE the last request to continue after getting the token
+            spider.last_request = request
```
jpmckinney: Isn't it possible for multiple simultaneous requests to meet the expiry condition? If so, should this be an array or requests to re-perform?

kingfisher_scrapy/spiders/paraguay_base.py
```diff
@@ -48,10 +49,9 @@ def from_crawler(cls, crawler, *args, **kwargs):
         return spider
 
     def start_requests(self):
-        # Start request access token
-        self.request_access_token()
         yield scrapy.Request(
             self.base_page_url,
+            dont_filter=True,
```
jpmckinney: Can you add a comment explaining in what case we might send duplicate requests?

jpmckinney:
It's not part of the PR, but I was very confused by what this line was trying to do:

```python
if 'auth' in request.meta and request.meta['auth'] is not None and not request.meta['auth']:
```

It would be clearer for this to be:

```python
if 'auth' in request.meta and request.meta['auth'] is False:
```

kingfisher_scrapy/spiders/paraguay_hacienda.py
aguilerapy: I think we can remove it from both scrapers.

kingfisher_scrapy/middlewares.py
aguilerapy: At this time, we defined `'CONCURRENT_REQUESTS': 1` in `custom_settings`, and that multiple expiry case couldn't happen. If we want to have simultaneous requests, the code must be analyzed for that.

kingfisher_scrapy/middlewares.py
```diff
@@ -88,4 +88,5 @@ def process_response(self, request, response, spider):
     @staticmethod
     def _expires_soon(spider):
         # spider MUST implement the expires_soon method
-        return spider.expires_soon(datetime.now() - spider.start_time) if spider.start_time else True
+        return spider.expires_soon(datetime.now() - spider.start_time) \
+            if (spider.start_time and spider.access_token) else True
```
jpmckinney: Please avoid one-line conditional expressions. Multiline if/else statements are much easier to read.

aguilerapy:
Actually, with `dont_filter`, scrappy doesn't filter the request and doesn't be logged, `dont_filter` works in requests for access tokens and in the continuation of last_request saved. We have INFO: "Time_diff", "Requesting access token" and "New access token" for logs in those cases. @jpmckinney

yolile:
@jpmckinney according to  [this](https://github.com/open-contracting/kingfisher-scrape/issues/323#issuecomment-594165048) comment we decided to keep using `crawler.crawl` but is ok to use this other solution that @aguilerapy found then?

jpmckinney:
I think either option is okay.

The option in this PR won't work if we removed `'CONCURRENT_REQUESTS': 1`. Do you know if `crawler.crawl` can work with concurrent requests? If not, they we can leave it as-is.

yolile:
> I think either option is okay.
> 
> The option in this PR won't work if we removed `'CONCURRENT_REQUESTS': 1`. Do you know if `crawler.crawl` can work with concurrent requests? If not, then we can leave it as-is.

Both options works only with CONCURRENT_REQUESTS set to 1

yolile:
@jpmckinney can we merge this?

jpmckinney:
Yes, it's approved :) I'm trying to fix Travis this week, so I'll merge in the meantime.

Issue #325, opened 2020-02-28
openopps: Add spider


jpmckinney:
See API documentation and credentials here: https://crm.open-contracting.org/issues/4572#note-9

Please keep the credentials secret, using environment variables.

aguilerapy:
Done in #326

Issue #322, opened 2020-02-24
Check use of .get() instead of []


jpmckinney:
Follow-up to #320 

If we expect a key to be set, then we should use `[]`, so that we get a KeyError if the response's structure changes. We should only ever use `.get()` if the key is not always present in the response. Otherwise, a crawl might not fail even if the response's structure changed.

Unlike other code, it is a good thing for a spider to be very picky about its input and to fail easily – because we don't have an easy way to test that the response structure is the same as before.

jpmckinney:
I think france and indonesia_bandung have a few more. The others are either followed by an if-statement or have a default value, so I think those are fine.

PR #326, opened 2020-03-03
Add OpenOpps scraper

aguilerapy:
#325


aguilerapy:
Credentials are kept secret, using environment variables, `KINGFISHER_OPENOPPS_USERNAME` and `KINGFISHER_OPENOPPS_PASSWORD`

aguilerapy:
Packages are saved as `release_package_list` for each [response](https://api.openopps.com/api/ocds/?format=json&ordering=releasedate&page=7&page_size=1000&releasedate__gte=2020-02-08&releasedate__lte=2020-02-08) in order to have fewer files.

kingfisher_scrapy/middlewares.py
```diff
@@ -84,3 +84,23 @@ def process_response(self, request, response, spider):
...
+        logging.info('Initialized authentication middleware with spider: {}.'.format(spider.name))
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.spider)
+
+    @staticmethod
+    def process_request(request, spider):
+        if 'token_request' in request.meta and request.meta['token_request']:
+            logging.info('Requesting access token...')
```
jpmckinney: Just curious – does the Scrapy log not include the access token request?

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class OpenOpps(BaseSpider):
+    name = 'open_opps'
+
+    access_token = None
+    api_limit = 10000  # OpenOpps API limit for search results
+    page_size = 1000
+    page_format = 'json'
+    ordering = 'releasedate'
```
jpmckinney: If we only use these variables in one place, I think it's better to put the values there, rather than have lots of class variables.

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+                'Authentication failed. Status code: {}. {}'.format(response.status, response.text))
+            raise AuthenticationFailureException()
+
+    def start_requests_pages(self):
+        # use larger ranges for filters with less than (api_limit) search results
+        if not self.sample:
+            release_date_gte_list = ['', '2009-01-01', '2010-01-01', '2010-07-01']
+            release_date_lte_list = ['2008-12-31', '2009-12-31', '2010-06-30', '2010-12-31']
+
+            for i in range(0, 4):
```
jpmckinney: To avoid a bug if we change the lists above, use `range(len(release_date_gte_list))` instead of `range(0, 4)`.

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+                        self.ordering,
+                        self.page_size,
+                        release_date_gte,
+                        release_date_lte
+                    ),
+                    headers={"Accept": "*/*", "Content-Type": "application/json"},
+                    meta={"release_date": date},
+                )
+
+                if self.sample:
```
jpmckinney: The sample logic is a bit complicated, so please add a comment like "The sample will be data released on the first day of 2011." (assuming I understood the logic)

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def parse(self, response):
+        if response.status == 200:
+            results = json.loads(response.text)
+            count = results.get('count')
+            if count <= 10000:
+
+                # data type changed to release package list in order to have fewer files
+                all_data = {}
+                for data in results.get('results'):
```
jpmckinney: Why use `.get()` instead of `[]`? See #322

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+                    json_data = data.get('json')
+                    if json_data:
+                        all_data.update({len(all_data): json_data})
```
jpmckinney: Why is this a dict and not a list?

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+from datetime import datetime, timedelta
+from math import ceil
+
+import scrapy
+
+from kingfisher_scrapy.base_spider import BaseSpider
+from kingfisher_scrapy.exceptions import AuthenticationFailureException
+
+
+class OpenOpps(BaseSpider):
```
jpmckinney: Please always add a comment with a link to the API's documentation.

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+                    )
+            else:
+                if 'release_date' in response.request.meta and response.request.meta['release_date']:
+                    parts = int(ceil(count / self.api_limit))  # parts we split a search that exceeds the limit
+                    split_h = int(ceil(24 / parts))  # hours we split
+                    date = response.request.meta['release_date']
+                    date = datetime.strptime(date, "%Y-%m-%d")
+
+                    # create time lists depending on how many hours we split a day
+                    start_hour_list = [(date + timedelta(hours=H)
```
jpmckinney: Why is `H` capitalized?

kingfisher_scrapy/spiders/open_opps.py
```diff
@@ -0,0 +1,181 @@
...
+
+                next_url = results.get('next')
+                if next_url:
+                    yield scrapy.Request(
+                        url=next_url,
+                        headers={"Accept": "*/*", "Content-Type": "application/json"},
+                    )
+            else:
+                if 'release_date' in response.request.meta and response.request.meta['release_date']:
+                    parts = int(ceil(count / self.api_limit))  # parts we split a search that exceeds the limit
```
jpmckinney: I understand the desire to minimize requests, but if any part still has too many results (e.g. if the results are concentrated within certain hours), then this will result in infinite requests, because it will constantly try (and fail) to split the day into the same number of parts.

kingfisher_scrapy/middlewares.py
aguilerapy: Scrapy only prints `"DEBUG: Crawled (200) <POST https://api.openopps.com/api/api-token-auth/>"`, I added `"Requesting access token..."` and `"New access token"` info to know when we request and get the token.

kingfisher_scrapy/spiders/open_opps.py
aguilerapy: We use these variables in `start_requests_pages` and `parse`, but now I put only in `start_requests_pages` affecting `parse` too.

kingfisher_scrapy/spiders/open_opps.py
aguilerapy: Changed to list.

kingfisher_scrapy/spiders/open_opps.py
aguilerapy: Changed to h.

kingfisher_scrapy/spiders/open_opps.py
aguilerapy: Fixed, we can now split a range of X `hours` into N `parts` of Y `hours` if still has too many results with recursion. Thanks for find that!

kingfisher_scrapy/middlewares.py
jpmckinney: I think Scrapy's message is enough, instead of also logging "Requesting access token...".

jpmckinney:
Thanks for the changes. I see I jumped ahead of @romifz so once Romi has reviewed (I haven't checked the new logic around splitting hours of the day), feel free to merge.

aguilerapy:
Minimum range of one hour: If we exceed the results in a range of one hour, we save the first 10,000 results of each search.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -0,0 +1,193 @@
...
+                next_url = results.get('next')
+                if next_url:
+                    yield scrapy.Request(
+                        url=next_url,
+                        headers={"Accept": "*/*", "Content-Type": "application/json"},
+                        meta={"release_date": release_date, "search_h": search_h},
+                    )
+            else:
+                # Change search filter if count exceeds the API limit or search_h > 1 hour
+                parts = int(ceil(count / self.api_limit))  # parts we split a search that exceeds the limit
```
romifz: What happens here if parts > 24?

aguilerapy: Then `split_h = 1` and the next request is `search_h = split_h = 1`, and if so, just save the first 10,000 results.

Only `search_h` and `split_h` are used to iterate ranges of hours, those variables will never be out of the 24-hour range.

jpmckinney:
Merging based on resolved comments – I haven't re-read the PR.

Issue #328, opened 2020-03-04
Use save_response_to_disk instead of save_data_to_disk where possible


jpmckinney:
I think it's possible in:

* [x] kenya_makueni
* [x] nepal_dhangadhi
* [x] nepal_portal
* [x] nigeria_portal
* [x] uganda_releases
* [x] uruguay_base

aguilerapy:
Uganda API could return the package with empty release list, anyway I saved it.

PR #330, opened 2020-03-05
Save responses

aguilerapy:
Uganda API could return the package with empty release list, anyway I saved it. #328

### DIFF
--- a/kingfisher_scrapy/spiders/kenya_makueni.py
+++ b/kingfisher_scrapy/spiders/kenya_makueni.py
@@ -1,5 +1,4 @@
 import hashlib
-import json
 from math import ceil
 
 import scrapy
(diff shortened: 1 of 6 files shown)


jpmckinney:
Actually, taking a step back, in what case does it have an empty release list?

aguilerapy:
Examples for this empty list case:

- https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/award?fy=2014-2015&pde=375
- https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/contract?fy=2019-2020&pde=684
- https://gpp.ppda.go.ug/adminapi/public/api/open-data/v1/releases/award?fy=2019-2020&pde=450

jpmckinney:
Aha, so some public entities just don't have any contracts. I think it's fine to save them without filtering them out like in the PR.

jpmckinney:
Thanks for the review, @yolile. Please delete branches after merging, to keep GitHub tidy :)

yolile:
@jpmckinney sure! I thought that the branches were auto deleted after merged but it seems that is not the case, so I will delete it by myself the next time

jpmckinney:
I just double-checked, and I see that there is a "Automatically delete head branches" option, which I've now enabled, so we don't need to manually delete branches :)

Issue #327, opened 2020-03-04
Refactor common code for handling JSON data in ZIP files


jpmckinney:
Most of the spiders that use `zipfile` use it in exactly the same way. Please extract the common logic to a method in the base spider.

PR #332, opened 2020-03-06
Add parse_zipfile method

aguilerapy:
#327

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -115,7 +115,27 @@ class KingfisherSpiderMixin:
 # https://docs.python.org/3.8/library/functions.html#super
 # https://rhettinger.wordpress.com/2011/05/26/super-considered-super/
 class BaseSpider(KingfisherSpiderMixin, scrapy.Spider):
-    pass
+    def parse_zipfile(self, response, data_type):
+        """
+        Handling response with JSON data in ZIP files
+        """
+        if response.status == 200:
+            zip_file = ZipFile(BytesIO(response.body))
+            for finfo in zip_file.infolist():
+                data = zip_file.open(finfo.filename).read()
+                if finfo.filename.endswith('.json'):
+                    filename = finfo.filename
+                else:
+                    filename = finfo.filename + '.json'
+                yield self.save_data_to_disk(data, filename, data_type=data_type, url=response.request.url)
+
+        else:
+            yield {
+                'success': False,
+                'file_name': response.request.meta['kf_filename'],
+                'url': response.request.url,
+                'errors': {'http_code': response.status}
+            }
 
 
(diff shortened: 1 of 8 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -150,3 +152,28 @@ def parse_next_link(self, response, data_type):
                 'url': response.request.url,
                 'errors': {'http_code': response.status}
             }
+
+
+class ZipSpider(BaseSpider):
```
jpmckinney: I'm fine with just adding the method to BaseSpider, to avoid a proliferation of classes.

kingfisher_scrapy/base_spider.py
```diff
@@ -150,3 +152,28 @@ def parse_next_link(self, response, data_type):
...
+    def parse_zipfile(self, response, data_type):
+        """
+        Handling response with JSON data in ZIP files
+
```
jpmckinney: Remove extra newline.

kingfisher_scrapy/base_spider.py
```diff
@@ -150,3 +152,28 @@ def parse_next_link(self, response, data_type):
...
+        """
+        if response.status == 200:
+            zip_files = ZipFile(BytesIO(response.body))
```
jpmckinney: Change to singular `zip_file` as there is only one.

kingfisher_scrapy/base_spider.py
```diff
@@ -150,3 +152,28 @@ def parse_next_link(self, response, data_type):
...
+            for finfo in zip_files.infolist():
+                data = zip_files.open(finfo.filename).read()
+                if finfo.filename.endswith('.json'):
```
jpmckinney: In what case does it not end in `.json`?

kingfisher_scrapy/base_spider.py
```diff
@@ -150,3 +152,28 @@ def parse_next_link(self, response, data_type):
...
+                    filename = finfo.filename
```
jpmckinney: I noticed that some spiders prepended `response.request.meta['kf_filename']`. Does it matter?

kingfisher_scrapy/base_spider.py
aguilerapy: In the case of Portugal, it returns a file without extension, https://dados.gov.pt/s/resources/contratos-publicos-formato-ocds-informacao-registada-no-portal-base-em-www-base-gov-pt/20191128-111631/base2-pt-ocds-201911.zip

kingfisher_scrapy/base_spider.py
aguilerapy: I verified, there's no conflict in names for those cases, `finfo.filename` works.

Issue #334, opened 2020-03-24
Update honduras_portal scrapers


romifz:
* Release and Record packages are included inside another JSON object, see http://www.contratacionesabiertas.gob.hn/api/v1/release/.
* Update the URL to use http://www.contratacionesabiertas.gob.hn/

PR #336, opened 2020-03-24
Update Honduras portal scrapers

romifz:
ref #334


jpmckinney:
Should we do that parallelization now, or create a new issue?

yolile:
I think that is better to create another issue, I will do it

yolile:
Done in #337

Issue #343, opened 2020-03-30
Openopps scraper fails after two hours of requests


aguilerapy:
Re-authenticate before the token expires. Follow up CRM issue 4572.

jpmckinney:
Closed by #344

PR #344, opened 2020-03-31
Fix openopps scraper

aguilerapy:
Issue #343. I think the best way to kept parallelism and re-authenticate before token expires, is by checking the time on the spider and flag when we should request a new token. A token is requested every 60 minutes, it's not exact since the new token request is not executed immediately when queued, because other requests are already processed and await download (due to parallelism). What do you think?

### DIFF
--- a/kingfisher_scrapy/spiders/openopps.py
+++ b/kingfisher_scrapy/spiders/openopps.py
@@ -217,7 +217,9 @@ class OpenOpps(BaseSpider):
                               "search_h": split_h},  # new search range
                     )
         else:
-            if response.status == 500:
+            # Message for pages that exceed the 10,000 search results in the range of one hour
+            # These are pages with status 500 and 'page=11' in the URL request
+            if response.status == 500 and response.request.url.count("page=11"):
                 self.logger.info('Status: {}. Results exceeded in a range of one hour, we save the '
                                  'first 10,000 data for: {}'.format(response.status, response.request.url))
             else:
--- a/kingfisher_scrapy/spiders/openopps.py
+++ b/kingfisher_scrapy/spiders/openopps.py
@@ -20,9 +20,6 @@ class OpenOpps(BaseSpider):
 
     access_token = None
     api_limit = 10000  # OpenOpps API limit for search results
-    request_time_limit = 60  # in minutes
-    re_authenticate = False  # flag for request a new token
-    start_time = None
 
     base_page_url = \
         'https://api.openopps.com/api/ocds/?' \
@@ -54,8 +51,6 @@ class OpenOpps(BaseSpider):
             method='POST',
             headers={"Accept": "*/*", "Content-Type": "application/json"},
             body=json.dumps({"username": self.username, "password": self.password}),
-            # send duplicate requests when we re-authenticate before the token expires
-            dont_filter=True,
             meta={'token_request': True},
             callback=self.parse_access_token
         )
@@ -67,11 +62,8 @@ class OpenOpps(BaseSpider):
             if token:
                 self.logger.info('New access token: {}'.format(token))
                 self.access_token = 'JWT ' + token
-                self.start_time = datetime.now()
-                if not self.re_authenticate:
-                    """ Start requests """
-                    return self.start_requests_pages()
-                self.re_authenticate = False
+                """ Start requests """
+                return self.start_requests_pages()
             else:
                 self.logger.error(
                     'Authentication failed. Status code: {}. {}'.format(response.status, response.text))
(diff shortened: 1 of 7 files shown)


aguilerapy:
I think some #342 commits got in here. Sorry.

jpmckinney:
Can you cherry-pick 292b68b onto a new branch? Or, @yolile can review and merge #342, and then we can rebase this PR.

aguilerapy:
The PR is now fixed! @jpmckinney

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -62,8 +67,11 @@ def parse_access_token(self, response):
             if token:
                 self.logger.info('New access token: {}'.format(token))
                 self.access_token = 'JWT ' + token
-                """ Start requests """
-                return self.start_requests_pages()
+                self.start_time = datetime.now()
+                if not self.re_authenticate:
+                    """ Start requests """
+                    return self.start_requests_pages()
+                self.re_authenticate = False
```
jpmckinney, line 74: Let's see if I understand the logic.

At first, `re_authenticate = False` and `start_requests` is called, it yields a request whose callback is `parse_access_token`, which returns `start_requests_pages`. (I don't know why it returns instead of just calling, since that method doesn't have a return value? It'd be clearer to have `return` on a separate line if the return value is not meaningful.)

`start_requests_pages` then yields requests, whose callback is implicitly `parse`. If time runs out, `parse` sets `re_authenticate = True` and yields a request whose callback is `parse_access_token`. Since `re_authenticate` is true, `parse_access_token` doesn't call `start_requests_pages`, then resets `re_authenticate` to false.

Meanwhile, the requests yielded by `start_requests_pages` eventually go through the middleware, which adds the access token. Critically, the middleware's callback is only called when the request is about to be made, so it will have the most recently-set token. (This isn't thread-safe, but that's okay because this part of Scrapy isn't threaded as I understand?.)

I think part of my confusion was that `re_authenticate` is being used to mean both whether we need to perform the initial authentication, and whether we are in the process of reauthenticating. If we change the logic in `parse_access_token` to:

```python
if self.initial_authentication:
    self.initial_authentication = False
    self.start_requests_pages()
    return
```

And then rename `re_authenticate` to `reauthenticating`, we'll have separated the logic and made the code easier to read.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -191,8 +215,13 @@ def parse(self, response):
...
                     )
         else:
-            yield {
-                'success': False,
-                'url': response.request.url,
-                'errors': {'http_code': response.status}
-            }
+            if response.status == 500:
+                self.logger.info('Status: {}. Results exceeded in a range of one hour, we save the '
+                                 'first 10,000 data for: {}'.format(response.status, response.request.url))
```
jpmckinney, line 220: I don't understand the error message. What causes the error? Also, can't there be a 500 status for all kinds of errors? Should we more narrowly test for the specific type of error?

kingfisher_scrapy/spiders/openopps.py
aguilerapy: Yes, so, `parse_access_token()` is the callback function, when it's called, we are in [step 6](https://docs.scrapy.org/en/latest/topics/architecture.html#data-flow), specifically in the spider, at that moment spider must return Request, BaseItem, Dict or None. 

With `return self.start_requests_pages()` in `parse_access_token()` we are queuing the requests because `start_requests_pages()` yields requests. Scrapy loses the trace if we delete the return of the callback function and with `self.start_requests_pages()`, Scrapy doesn't go to the function we have defined `def start_requests_pages (self):`

Yes, the middleware callback is only called when the request is in step 4. Scrapy takes [16 requests](https://docs.scrapy.org/en/latest/topics/settings.html#concurrent-requests) (default) from the queue and processes them by placing tokens on each one. I don't know exactly how the [Downloader](https://docs.scrapy.org/en/latest/topics/architecture.html#downloader) works below, but we make sure to place valid tokens by restarting them every 60 minutes, away from the API's 2-hour limit.

I will change some things so that it's easier to read!

kingfisher_scrapy/spiders/openopps.py
aguilerapy: Causes by the pages that exceed the 10,000 search in the range of one hour. We cut here because the code becomes very complex if we split ranges of minutes and seconds. 

You're right, I'm going to correct to capture only the cases where we want to display the message.

kingfisher_scrapy/spiders/openopps.py
jpmckinney: Aha, yes, you're right, we need `return self.start_requests_pages()`, because it returns a generator of Requests.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -191,8 +217,15 @@ def parse(self, response):
...
                     )
         else:
-            yield {
-                'success': False,
-                'url': response.request.url,
-                'errors': {'http_code': response.status}
-            }
+            # Message for pages that exceed the 10,000 search results in the range of one hour
+            # These are pages with status 500 and 'page=11' in the URL request
+            if response.status == 500 and response.request.url.count("page=11"):
```
yolile, line 222: @aguilerapy will it always be "page=11"?

aguilerapy: Yes, because on each page there are a maximum of 1,000 pages, and if we get to page 11 we exceed 10,000 search results.

jpmckinney:
I'm not sure why the PR checks didn't report back – hopefully they'll work for new PRs.

PR #345, opened 2020-03-31
requirements: Run pip-tools to upgrade all

odscjames:
This
* adds rarfile to dev locked dependencies
* upgrades pip-tools to a version that works with the latest version of pip
* fixes Twisted security alert


romifz:
I tested this locally and works with no issues.

jpmckinney:
Superseded by other commits.

Issue #350, opened 2020-04-06
Add spider arguments to OpenOpps


aguilerapy:
Add spider arguments to control the date range to download. Follow up CRM issue 4572.

jpmckinney:
While doing this, you can also consider #346

aguilerapy:
Closed  #351

Issue #346, opened 2020-04-03
Implement from_date and until_date spider arguments


jpmckinney:
Remember to document new spider arguments in the spider's docstring.

Add the filters to these spiders:

* [x] honduras_oncae (from_date)
* [x] honduras_oncae (until_date)
* [x] honduras_portal_bulk_files (from_date)
* [x] honduras_portal_bulk_files (until_date)
* [x] paraguay_dncp_records (until_date)
* [x] paraguay_dncp_releases (until_date)
* [x] scotland_public_contracts (until_date)

The following are periodic spiders without documentation of `from_date` and `until_date`:

* [x] moldova_old
* [x] nepal_portal
* [x] uruguay_historical
* [x] uruguay_records
* [x] uruguay_releases

The following sources don't support date filters:

* honduras_portal_records
* honduras_portal_releases

The following sources can be checked. The assumption is that they don't support date filters:

* armenia (skipping, too complex)
* [x] costa_rica_poder_judicial_releases
* [ ] paraguay_hacienda
* [x] spain_zaragoza
* [x] afghanistan_records
* [x] afghanistan_releases
* [x] argentina_buenos_aires
* [x] australia
* [x] bolivia_agetic
* [x] canada_buyandsell
* [x] chile_compra_bulk
* [x] colombia_bulk
* [x] costa_rica_poder_judicial_records
* [x] dominican_republic
* [x] france
* [x] georgia_opendata
* [x] georgia_records
* [x] georgia_releases
* [x] italy
* [x] kenya_makueni
* [x] kyrgyzstan
* [x] malta
* [x] mexico_administracion_publica_federal
* [x] mexico_grupo_aeroporto
* [x] mexico_inai
* [x] mexico_inai_portal
* [x] mexico_nuevo_leon_records
* [x] mexico_nuevo_leon_releases
* [x] mexico_quien_es_quien
* [x] moldova
* [x] moldova_mtender
* [x] nepal_dhangadhi
* [x] nigeria_budeshi_records
* [x] nigeria_budeshi_releases
* [x] nigeria_portal
* [x] portugal
* [x] uganda_releases
* [x] uk_contracts_finder
* [x] uk_fts
* [x] zambia

~~If any filters are applied, they should be logged, so that we can identify whether a collection is complete or incomplete. (#112).~~ (Done in BaseSpider for both recognized spider arguments and kwargs in general)

If the API doesn't support the filters, we can consider performing the filtering in the item pipeline (see #277).

[...]

PR #351, opened 2020-04-06
Add spider arguments to OpenOpps

aguilerapy:
Changes for #350 and some implementations for #346.


kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
 class OpenOpps(BaseSpider):
     """
     API documentation
+      -  https://docs.google.com/document/d/1u0da3BTU7fBFjX6i7j_tKXa1YwdXL7hY4Kw9GdsaAr0/edit
```
jpmckinney, line 15: You can omit the `- ` and it will render as a definition list (instead of a bullet list inside a definition list).

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
+    Swagger API documentation
       -  https://api.openopps.com/api/ocds/
```
jpmckinney, line 17: This is the actual API URL. We can omit it and just have the Swagger docs.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
+      KINGFISHER_OPENOPPS_USERNAME
+        Username variable for credentials.
+      KINGFISHER_OPENOPPS_PASSWORD
+        Password variable for credentials.
     """
     name = 'openopps'
 
     access_token = None
     api_limit = 10000  # OpenOpps API limit for search results
+    download_delay = 0.9
```
jpmckinney, line 38: Why 900ms?

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -80,46 +96,80 @@ def start_requests_pages(self):
...
+                        self.until_date = datetime.now().strftime("%Y-%m-%d")
+                    start_date = datetime.strptime(self.from_date, "%Y-%m-%d")
+                    end_date = datetime.strptime(self.until_date, "%Y-%m-%d")
+                    date_list = [(start_date + timedelta(days=d)).strftime("%Y-%m-%d")
+                                 for d in range((end_date - start_date).days + 1)]
+
+                except ValueError as e:
+                    self.logger.error(e.args)
+                    self.logger.info('See API documentation for date format.')
+                    return
```
jpmckinney, line 126: Let's put date parsing logic in the base spider, so it can be re-used. We can validate the arguments on initialization.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -80,46 +96,80 @@ def start_requests_pages(self):
...
+                    release_date_gte = date
+                    release_date_lte = date
+                    yield scrapy.Request(
+                        url=self.base_page_url.format(
+                            release_date_gte,
+                            release_date_lte
+                        ),
+                        headers={"Accept": "*/*", "Content-Type": "application/json"},
+                        meta={"release_date": date, "search_h": search_h},
+                    )
```
jpmckinney, line 138: Instead of copying this code (plus the `date_list` assignment) from below, you can extract it to a new method (using `yield from`).

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -80,46 +96,80 @@ def start_requests_pages(self):
...
+        if self.sample:
+            date = '2011-01-01'
+            yield scrapy.Request(
+                url=self.base_page_url.format(
+                    date,
+                    date,
+                ),
+                headers={"Accept": "*/*", "Content-Type": "application/json"},
+                meta={"release_date": date, "search_h": search_h},
+            )
```
jpmckinney, line 109: If we extract a method as in another comment, we can use it again here.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -80,46 +96,80 @@ def start_requests_pages(self):
         search_h = 24  # start splitting one day search
         self.base_page_url = self.base_page_url.format(page_format, ordering, page_size, '{}', '{}')
```
jpmckinney, line 97: Instead of formatting with `'{}'`, you can just escape the original `{}` that you don't want to replace this first time with `{{}}`.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
     """
     API documentation
+      -  https://docs.google.com/document/d/1u0da3BTU7fBFjX6i7j_tKXa1YwdXL7hY4Kw9GdsaAr0/edit
+    Swagger API documentation
       -  https://api.openopps.com/api/ocds/
       -  https://api.openopps.com/api/schema/
-      -  https://docs.google.com/document/d/1u0da3BTU7fBFjX6i7j_tKXa1YwdXL7hY4Kw9GdsaAr0/edit
+    Spider arguments
+      sample
+        Download only data released on 2011-01-01 for sample.
```
jpmckinney, line 21: > Download only data released on 2011-01-01.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
+      from_date
+        The start date for range to download. See API documentation for date format.
+        Date '2011-01-01' by default if until_date is set and from_date not.
```
jpmckinney, line 24: We're parsing the dates as `%Y-%m-%d`, so we can be explicit:

```rst
Download only data from this date onward (YYYY-MM-DD format). If ``until_date`` is provided, defaults to '2011-01-01'.
```

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
+      until_date
+        The end date for range to download. See API documentation for date format.
+        Today's date by default if from_date is set and until_date not.
```
jpmckinney, line 27:
```rst
Download only data until this date (YYYY-MM-DD format). If ``from_date`` is provided, defaults to today.

File: kingfisher_scrapy/spiders/openopps.py [hunk 11]
```diff
@@ -12,14 +12,30 @@
...
+    Environment variables
+      KINGFISHER_OPENOPPS_USERNAME
+        Username variable for credentials.
```
jpmckinney, line 30: As documentation, it'd be more useful to say where they can get an account.

```rst
To get an API account, contact contact@openopps.com.
```

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -12,14 +12,30 @@
...
+      KINGFISHER_OPENOPPS_PASSWORD
+        Password variable for credentials.
```
jpmckinney, line 32: > Your API account password.

kingfisher_scrapy/spiders/openopps.py
aguilerapy: I think it seems like a reasonable time to avoid hitting the server too hard, could be changed.

jpmckinney: Let's just make it 1. When I see a number like 0.9, I assume there's a specific reason to be shaving off 100ms :)

aguilerapy:
There were some warnings for Paraguay scrapers.

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -73,53 +88,63 @@ def parse_access_token(self, response):
...
+
+                for i in range(len(release_date_gte_list)):
+                    yield scrapy.Request(
+                        url=self.base_page_url.format(
+                            release_date_gte_list[i],
+                            release_date_lte_list[i]
+                        ),
+                        headers={"Accept": "*/*", "Content-Type": "application/json"},
+                        meta={"release_date": release_date_gte_list[i], "search_h": search_h},
+                    )
```
jpmckinney, line 140: To avoid some repetition, we can have a method that accepts a start_date, end_date and search_h, and returns a Request. This code (and the other occurrence) could then `yield self.new_method_name(start_date, end_date, search_h)`

kingfisher_scrapy/base_spider.py
```diff
@@ -137,6 +153,30 @@ def parse_zipfile(self, response, data_type):
...
+            if not self.until_date:
+                self.until_date = datetime.now().strftime(date_format)
+            from_date = datetime.strptime(self.from_date, date_format)
+            until_date = datetime.strptime(self.until_date, date_format)
+            return from_date, until_date
+
+        except ValueError as e:
+            self.logger.error(e.args)
+            self.logger.info('See API documentation for date format.')
+            return False, False
```
jpmckinney, line 178: I think the crawl should abort (`raise scrapy.exceptions.CloseSpider()`) if the parameters can't be parsed. Presently, I think the crawl will just write to the log, but it'll otherwise appear to "succeed" (`finish_reason='finished'`) in terms of crawl status (instead of `finish_reason='cancelled'` if the exception is raised).

kingfisher_scrapy/spiders/openopps.py
```diff
@@ -73,53 +88,63 @@ def parse_access_token(self, response):
...
+        # Case if we want to download a sample
+        if self.sample:
+            date = datetime(2011, 1, 1)
+            yield from self.request_date_list(date, date, search_h)
+        else:
+            # Case if we have date range parameters
+            if self.from_date or self.until_date:
+                start_date, end_date = self.parse_date_arguments(
+                    date_format='%Y-%m-%d',
+                    default_from_date='2011-01-01'
```
jpmckinney, line 123: Instead of calling `parse_date_arguments` from the specific spider, we should call it from the base spider's `__init__`. We'll just need a way for specific spiders to set their own defaults (perhaps with class-level variables).

aguilerapy: Is there no problem calling `scrapy.exceptions.CloseSpider()` from the base spider's `__init__`?

jpmckinney: Let's try it locally and see :)

jpmckinney: Ideally, we can have a test to ensure it works.

aguilerapy: imagen

When I call `scrapy.exceptions.CloseSpider()` from the base spider's `__init__`, that exception is not treated because openopps spider has not been created yet, it's okay to show this anyway?

jpmckinney: What if we put the date logic in a small middleware's `from_crawler` method? https://docs.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.from_crawler

jpmckinney: Nevermind – CloseSpider is only checked in response callbacks. I'm not sure what the Scrapy best practice here is. `NotConfigured` is also not the right exception, because the crawl will continue with the middleware disabled. We can maybe try `UsageError` (might have prettier output), or create our own exception class.

aguilerapy: I call `parse_date_arguments` from the base spider's `from_crawler`, because at this point the specific spider class is created and we can access to the defaults attributes (`date_format` and `default_from_date`), what do you think?

aguilerapy: Also I create an `SpiderArgumentError` exception, that would be useful for any other spider argument.

kingfisher_scrapy/base_spider.py
```diff
@@ -100,13 +115,40 @@ def _get_crawl_path(self):
...
+        else:
+            self.message = None
+
+    def __str__(self):
+        if self.message:
+            return self.message
+        else:
+            return 'SpiderArgumentError has been raised.'
+
+
```
jpmckinney, line 131: Can you move this to `kingfisher_scrapy/exceptions.py` and have it inherit from `KingfisherScrapyError`?

Also, I think we can omit the `__init__` and `__str__` methods and just have the default Exception behavior.

kingfisher_scrapy/base_spider.py
```diff
@@ -129,6 +171,28 @@ def parse_zipfile(self, response, data_type):
                 'errors': {'http_code': response.status}
             }
 
+    def parse_date_arguments(self, date_format, default_from_date=None):
+        """
+        Parsing logic for date range to download.
+        Specific spider class must have the attributes ``date_format`` and optionally ``default_from_date``.
```
jpmckinney, line 177: Instead of having a different date format for each spider, let's just always use the same format. It'll be simpler for users to only have to learn one format.

kingfisher_scrapy/base_spider.py
```diff
@@ -129,6 +171,28 @@ def parse_zipfile(self, response, data_type):
...
+        If ``until_date`` is provided, defaults to ``default_from_date``.
+        If ``from_date`` is provided, defaults to today.
+
+        Returns datetime objects or false if the format is wrong.
+        """
+        try:
+            if not self.from_date:
+                self.from_date = default_from_date
+            if not self.until_date:
+                self.until_date = datetime.now().strftime(date_format)
```
jpmckinney, line 189: This can go outside the `try` as it won't cause a ValueError.

kingfisher_scrapy/base_spider.py
```diff
@@ -129,6 +171,28 @@ def parse_zipfile(self, response, data_type):
...
+            self.from_date = datetime.strptime(self.from_date, date_format)
+            self.until_date = datetime.strptime(self.until_date, date_format)
+
+        except ValueError as e:
+            raise SpiderArgumentError(e.args[0] + ", see API documentation.")
```
jpmckinney, line 194: I don't think users will know where to look for "see API documentation" so let's just do:

```python
raise SpiderArgumentError(e)
```

which will cause:

```
ValueError: time data 'the-user-input' does not match format '%Y-%m-%d'
```

to be written to the log/stderr, which is clear enough.

kingfisher_scrapy/base_spider.py
```diff
@@ -15,17 +16,31 @@ class KingfisherSpiderMixin:
...
+
+    .. code:: bash
+
+        scrapy crawl spider_name -a from_date='2010-01-01'
+
+    Set the end date for range to download:
+
+    .. code:: bash
+
+        scrapy crawl spider_name -a until_date='2020-01-01'
```
jpmckinney, line 29: We can omit the single quotes around the dates, so that users have less to type, and so that they don't accidentally omit a quote.

kingfisher_scrapy/base_spider.py
```diff
@@ -100,13 +115,40 @@ def _get_crawl_path(self):
...
+
+
 # `scrapy.Spider` is not set up for cooperative multiple inheritance (it doesn't call `super()`), so the mixin must be
 # the first declared parent class, in order for its `__init__()` and `from_crawler()` methods to be run.
 #
 # https://github.com/scrapy/scrapy/blob/1.8.0/scrapy/spiders/__init__.py#L25-L32
 # https://docs.python.org/3.8/library/functions.html#super
 # https://rhettinger.wordpress.com/2011/05/26/super-considered-super/
 class BaseSpider(KingfisherSpiderMixin, scrapy.Spider):
+    @classmethod
```
jpmckinney, line 139: I'm a little concerned about adding methods to `BaseSpider` that in principle should be shared with `BaseXMLFeedSpider` (which is what `KingfisherSpiderMixin` is for).

However, since no spiders use `BaseXMLFeedSpider` anymore, maybe we should just remove it, and fold `KingfisherSpiderMixin` into `BaseSpider`.

If we do that, then we can remove the comments above BaseSpider about multiple inheritance.

kingfisher_scrapy/base_spider.py
```diff
@@ -129,6 +171,28 @@ def parse_zipfile(self, response, data_type):
                 'errors': {'http_code': response.status}
             }
 
+    def parse_date_arguments(self, date_format, default_from_date=None):
```
jpmckinney, line 174: Since this is a very small method that is only called from one place, I prefer to in-line it in `from_crawler`. We can remove the docstring since it's repetitive with the individual spider docstring. (If we kept it, we'd have to fix the return description, etc.).

kingfisher_scrapy/base_spider.py
aguilerapy: I understand that each API can handle a different format, right?

The idea is to kept one format and change the input format to the correct one for each API?

jpmckinney: A Scrapy user should just have to learn one format, which gets based to a Python datetime. 

We can then re-format that datetime as needed in each spider before sending the request to an API.

kingfisher_scrapy/base_spider.py
aguilerapy: #359 for solve this.

kingfisher_scrapy/base_spider.py
yolile: @jpmckinney the date_format is now just year, month and date, but in some cases, some APIs, as Paraguay, also support time, should I update the date_format to include the time? eg. %Y-%m-%d %H:%M:%S

jpmckinney: Hmm, we don't want to require people to add time components that have no effect.

Let's instead support both with and without time component. Please open a new issue if you want to discuss further, as we shouldn't discuss changes in functionality in a merged PR :)

PR #353, opened 2020-04-07
Update paraguay dncp spider according to the new URL

yolile:


yolile:
@jpmckinney can we skip de travis check here too?

jpmckinney:
Sure. I'm fixing OCDS profiles first. Then I'll do software repos.

PR #356, opened 2020-04-15
Remove HttpProxyWithSpiderArgsMiddleware

### LINKED ISSUE #296: Check whether HttpProxyWithSpiderArgsMiddleware is needed
~~My understanding is that Scrapy's [HttpProxyMiddleware](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpproxy) is [enabled by default](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#httpproxy-enabled), so we don't need `HttpProxyWithSpiderArgsMiddleware`.~~

* [x] If we close #293, then we can also remove `spider.proxies` in `paraguay_base.py`.

* [x] In PR #271, I make the `http_proxy` and `https_proxy` spider arguments easier to access. If we end up removing `HttpProxyWithSpiderArgsMiddleware`, we can remove these, too.

jpmckinney:
closes #296

@aguilerapy Can you test that it uses proxies when running something like this:

```
env http_proxy=YOUR-PROXY-URL https_proxy=YOUR-PROXY-URL scrapy crawl spider_name -s HTTPPROXY_ENABLED=True
```

And that it doesn't use proxies when `-s HTTPPROXY_ENABLED=True` is omitted (even if the environment variables are set)?

I also noticed that `paraguay_dncp_base` sets the priority of `ParaguayAuthMiddleware` to 543 (which is the same value as in the [Scrapy example](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#activating-a-downloader-middleware)), but it had set the priority of `HttpProxyWithSpiderArgsMiddleware` to 350, whereas Scrapy's `HttpProxyMiddleware` has a priority of 750. Do we need to change the priority of `ParaguayAuthMiddleware`? Here are all the default priorities: https://docs.scrapy.org/en/latest/topics/settings.html#std:setting-DOWNLOADER_MIDDLEWARES_BASE

### DIFF
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -58,6 +58,11 @@ To run a spider (that is, to start a "crawl"), replace ``spider_name`` below wit
 
     scrapy crawl spider_name
 
+.. _sample:
+
+Download a sample
+~~~~~~~~~~~~~~~~~
+
 To download only a sample of the available data, add the ``sample=true`` spider argument:
 
 .. code-block:: bash
@@ -68,8 +73,8 @@ Scrapy will then output a log of its activity.
 
 .. _proxy:
 
-Using an HTTP proxy
-~~~~~~~~~~~~~~~~~~~
+Use a proxy
+~~~~~~~~~~~
 
 .. note::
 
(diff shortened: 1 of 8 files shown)


jpmckinney:
Before deploying, we need to close https://github.com/open-contracting/deploy/pull/127

I think there are some failures due to flake8 issues that are fixed in #351

aguilerapy:
> env http_proxy=YOUR-PROXY-URL https_proxy=YOUR-PROXY-URL scrapy crawl spider_name -s HTTPPROXY_ENABLED=True
> 
Yes, HttpProxyMiddleware are enabled when running this.
> 
> And that it doesn't use proxies when `-s HTTPPROXY_ENABLED=True` is omitted (even if the environment variables are set)?
> 
HttpProxyMiddleware are disabled omitting `-s HTTPPROXY_ENABLED=True` even if the environment variables are set.

I'm going to review the priority section.

aguilerapy:
I think there is no need to change the priority. But I get 401 with POST https://contrataciones.gov.py/datos/api/v3/doc/oauth/token, I don't know if the `KINGFISHER_PARAGUAY_DNCP_REQUEST_TOKEN` configuration changed?

jpmckinney:
I just tried with my token and I get 200.

Issue #359, opened 2020-04-21
Remove BaseXMLFeedSpider


aguilerapy:
From @jpmckinney https://github.com/open-contracting/kingfisher-scrape/pull/351#discussion_r409170288

aguilerapy:
Closed #363.

PR #363, opened 2020-04-22
Remove BaseXMLFeedSpider

aguilerapy:
#359


aguilerapy:
I think we can extract `parse_json_lines` and `parse_zipfile` methods into a new `ZipSpider` class, since only some specific spiders use them, what you think @jpmckinney?

jpmckinney:
Hmm, up to you!

Issue #154, opened 2019-05-09
Buenos Aires: MemoryError


jpmckinney:
@romifz:

> in collection_file table, some files have ["MemoryError()"] in the errors column

@odscjames:

> Romi got it to work locally by using Flask with the dev server option. I wonder if it's UWSGI then? But if it was, that's not how I would expect that error to appear.

romifz:
This also happens with buenos-aires collection (@yolile).

By the way, shouldn't this issue be on the kingfisher-process project?

jpmckinney:
I don't know - is the MemoryError occurring during a crawl, and then getting sent to kingfisher-process' API? If so, I think the fix is in kingfisher-scrape.

romifz:
Apparently there is no errors sent during a crawl, see [this log](http://scrape.ocdskingfisher.opendataservices.coop/logs/kingfisher/honduras_oncae/73db0aca6a8b11e981f4b42e99482237.log). API calls are not returning error codes neither (I've seen HTTP error codes when sending files with incorrect formats).

robredpath:
@romifz is this something that you'd like someone like @odscjames to look into? How urgent is it?

romifz:
It's not urgent, for now I've been downloading the files using kingfisher-scrape locally and loading them to kingfisher-process using local-load.

odscjames:
When I run this locally on Django debug server I report a memory useage of 681 MB for processing the largest file. UWSGI is currently capped at 500MB. (Tho I am puzzled this isn't the memory error I've seen before from UWSGI)

Anyway, the short term fix is simply to increase memory. I think that server has 64GB of RAM after all.

The long term fix would be to use the Redis que - when a large file comes in, the web process simply stores it, returns a response to the caller as quick as possible and puts a message on the que. A Redis worker then processes it - they have no memory limits, I think, and they can take a longer time to process larger files without an HTTP agent waiting for a response.

odscjames:
Limit now increased to 1GB. We also have a 10 min limit per request, but we haven't seen that being broken yet.

romifz:
@odscjames I checked [Buenos Aires file](http://cdn.buenosaires.gob.ar/datosabiertos/datasets/buenos_aires_compras_open_contracting/20190416_bsas_release.zip) and it has ~800MB (uncompressed), so the memory limit should work for now! This file gets updated every 15 days, so it will grow with time,  but I think it would take a while to reach the 1GB cap.

[...]

PR #366, opened 2020-04-24
[WIP] Split large files with ijson

yolile:
ref #154


kingfisher_scrapy/spiders/argentina_buenos_aires.py
```diff
@@ -17,29 +14,25 @@ class ArgentinaBuenosAires(BaseSpider):
...
-                for finfo in zip_file.infolist():
-                    data = zip_file.open(finfo.filename).read()
-                    yield self.save_data_to_disk(data, finfo.filename, url=response.request.url,
-                                                 data_type='release_package')
+            data = json.loads(response.text)
+            for resource in data['result']['resources']:
+                if resource['format'].upper() == 'JSON':
+                    yield scrapy.Request(
+                        url=resource['url']
+                    )
```
jpmckinney, line 28: Can fit on one line now.

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
     def parse_zipfile(self, response, data_type, file_format=None, encoding='utf-8'):
         """
         Handling response with JSON data in ZIP files
 
         :param str file_format: The zipped files format. If this is set to 'json_lines', then the zipped file will be
-                                slitted by lines before send it to kingfisher-process and only the zip file will be
-                                stored as file.
+                                splitted by lines before send it to kingfisher-process and only the zip file will be
+                                stored as file. If it is set to 'release_package' the zipped file will be splitted by
+                                releases.
```
jpmckinney, line 237: Fixing some typos:

> The zipped file's format. If this is set to "json_lines", then each line of the zipped file will be yielded separately. If this is set to "release_package", then the releases will be re-packaged in groups of :const:`~kingfisher_scrapy.base_spider.BaseSpider. MAX_RELEASES_PER_PACKAGE` and yielded. In both cases, only the zipped file will be saved to disk. If this is not set, the file will be yielded and saved to disk.

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
+        def default(obj):
+            if isinstance(obj, Decimal):
+                return float(obj)
+            raise TypeError('%s is not JSON serializable' % repr(obj))
+
+        return json.dumps(data, default=default)
 
 
 class ZipSpider(BaseSpider):
+
```
jpmckinney, line 229: Remove extra newline.

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
+    @staticmethod
+    def json_dumps(data):
+        """
+        From ocdskit, returns the data as JSON.
+        """
+        def default(obj):
+            if isinstance(obj, Decimal):
+                return float(obj)
+            raise TypeError('%s is not JSON serializable' % repr(obj))
```
jpmckinney, line 223: Use `return json.JSONEncoder().default(obj)` instead.

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
+                yield from self._parse_json_item(packages, self.json_dumps(package), data_type, url, encoding)
+                package[array_field_name].clear()
+        if package[array_field_name]:
+            yield from self._parse_json_item(packages + 1, self.json_dumps(package), data_type, url, encoding)
+
+    @staticmethod
+    def json_dumps(data):
+        """
+        From ocdskit, returns the data as JSON.
+        """
```
jpmckinney, line 219: > Dumps JSON to a string, and returns it.

The OCDS Kit code has changed quite a bit (can dump generators, etc.).

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
+            if self.sample and number > self.MAX_SAMPLE:
+                break
+            if len(package[array_field_name]) < self.MAX_RELEASES_PER_PACKAGE:
+                package[array_field_name].append(item)
+            else:
+                packages += 1
+                yield from self._parse_json_item(packages, self.json_dumps(package), data_type, url, encoding)
+                package[array_field_name].clear()
+        if package[array_field_name]:
+            yield from self._parse_json_item(packages + 1, self.json_dumps(package), data_type, url, encoding)
```
jpmckinney, line 213: I think you can instead do (I haven't tested):

```python
from ocdskit.util import grouper

def parse_json_array(self, f_package, f_list, data_type, url, encoding='utf-8', array_field_name='releases'):
    if self.sample:
        size = self.MAX_SAMPLE
    else:
        size = self.MAX_RELEASES_PER_PACKAGE

    package = self.get_package(f_package, array_field_name)

    for number, items in enumerate(grouper(ijson.items(f_list, '{}.item'.format(array_field_name)), size), 1):
        package[array_field_name] = items
        yield self._build_file_item(number, self.json_dumps(package), data_type, url, encoding)
        if self.sample:
            break
```

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
...
+                    if value:
+                        package['extensions'].append(value)
+                elif '.' in prefix:
+                    object_name = prefix.split('.')[0]
+                    object_field = prefix.split('.')[1]
+                    if object_name not in package:
+                        package[object_name] = {}
+                    package[object_name][object_field] = value
+                else:
+                    package[prefix] = value
```
jpmckinney, line 194: This will work as long as the publisher only uses standard, package-level fields. Otherwise, if they have `{"extra":{"deep":{"property":1}}}`, it will not work.

If we want something more robust, we can re-implement [`ijson.common.items`](https://github.com/ICRAR/ijson/blob/master/ijson/common.py#L317) and [`ijson.common.items_basecoro`](https://github.com/ICRAR/ijson/blob/master/ijson/common.py#L137) (note: we should upgrade to ijson 3 as part of this, as it's internal APIs have changed compared to ijson 2), and just add a couple lines for it to `continue` if the prefix starts with the array name.

kingfisher_scrapy/base_spider.py
```diff
@@ -152,35 +158,89 @@ def _get_crawl_path(self):
             name += '_sample'
         return os.path.join(name, self.get_start_time('%Y%m%d_%H%M%S'))
 
+    @staticmethod
+    def _parse_json_item(number, line, data_type, url, encoding):
+        yield {
+            'success': True,
+            'number': number,
+            'file_name': 'data.json',
+            'data': line,
```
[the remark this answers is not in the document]
jpmckinney: Actually, let's fix that separately in #371, as CRM issue 5893 is waiting on it.

kingfisher_scrapy/base_spider.py
jpmckinney: Instead of importing ocdskit, you can just copy the grouper method over (it was itself copied from a Python cookbook).

kingfisher_scrapy/base_spider.py
```diff
@@ -152,18 +158,46 @@ def _get_crawl_path(self):
...
+    def parse_json_array(self, f_package, f_list, data_type, url, encoding='utf-8', array_field_name='releases'):
+        if self.sample:
+            size = self.MAX_SAMPLE
+        else:
+            size = self.MAX_RELEASES_PER_PACKAGE
+
+        package = self.get_package(f_package, array_field_name)
+
+        for number, items in enumerate(util.grouper(ijson.items(f_list, '{}.item'.format(array_field_name)), size), 1):
+            package[array_field_name] = [item for item in items if item is not None]
```
jpmckinney, line 198: In what case is an item None? You can also do `list(filter(None, items))`.

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
...
+
+from ijson import utils, ObjectBuilder
+
+
+@utils.coroutine
+def items_basecoro(target, prefix, map_type=None, array_name=None):
+    """
+    An couroutine dispatching native Python objects constructed from the events
+    under a given prefix.
+    """
```
jpmckinney, line 13: We can replace the original docstring with:

```
This is copied from ``ijson/common.py``. An ``array_name`` argument is added. If the ``array_name`` is in the current path, the current event is skipped. Otherwise, the method is identical.
```

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
...
+                    current, event, value = (yield)
+                del builder.containers[:]
+                target.send(builder.value)
+            else:
+                target.send(value)
+
+
+def items(events, prefix, map_type=None, array_name=None):
+    """Like ijson.items, but takes events generated via ijson.parse instead of
+    a file"""
```
jpmckinney, line 33: Replace with:

```
This is copied from ``ijson/common.py``. An ``array_name`` argument is added, which is passed as a keyword argument to :meth:`~kingfisher_scrapy.util.items_basecoro`. Otherwise, the method is identical.
```

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
...
+    return utils.coros2gen(events, (items_basecoro, (prefix, ), {'map_type': map_type, 'array_name': array_name}))
```
jpmckinney, line 34: Let's use the same whitespace as the original, to make it easier to compare visually:

```python
    return utils.coros2gen(events,
        (items_basecoro, (prefix,), {'map_type': map_type, 'array_name': array_name})
    )
```

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def default(obj):
+    """
+    From ocdskit, returns the data as JSON.
+    """
```
jpmckinney, line 40: Change to:

```
Dumps JSON to a string, and returns it.
```

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
...
+    if isinstance(obj, Decimal):
+        return float(obj)
+    try:
+        iter(obj)
+    except TypeError:
+        pass
```
jpmckinney, line 46: We need to either delete this, or replace it with:

```python
   try:
       iterable = iter(o)
   except TypeError:
       pass
   else:
       return list(iterable)
```

As-is, the code does nothing. If we replace it, we can do `filter(None, items)` from another comment instead of `list(filter(None, items))`

kingfisher_scrapy/base_spider.py
```diff
@@ -172,15 +206,17 @@ def parse_zipfile(self, response, data_type, file_format=None, encoding='utf-8')
         """
         Handling response with JSON data in ZIP files
 
-        :param str file_format: The zipped files format. If this is set to 'json_lines', then the zipped file will be
-                                slitted by lines before send it to kingfisher-process and only the zip file will be
-                                stored as file.
+        :param str file_format: The zipped file's format. If this is set to "json_lines", then each line of the zipped
+                                file will be yielded separately. If this is set to "release_package", then the releases
+                                will be re-packaged in groups of :const:~kingfisher_scrapy.base_spider.BaseSpider.
```
jpmckinney, line 211: Single backticks are missing around:

```
`~kingfisher_scrapy.base_spider.BaseSpider.MAX_RELEASES_PER_PACKAGE`
```

kingfisher_scrapy/util.py
```diff
@@ -0,0 +1,53 @@
+import itertools
+import json
+from decimal import Decimal
+
+from ijson import utils, ObjectBuilder
```
jpmckinney, line 5: Run `isort -w 119` to fix the test failure.

jpmckinney: Also, don't forget to add `ijson>=3` to requirements.in and run:

```
pip-compile; pip-compile requirements_dev.in
```

yolile: I got errors running `pip-compile` using pip 20.1, the current pip-compile version doesn't support pip>=20, the [5.5](https://github.com/jazzband/pip-tools/commit/d60d4169ec7030e89c7ae728b70366f83d662985#diff-4f8106c516356a37ed08defa42ad9a0d) version does. Besides, can we document these development process somewhere?

jpmckinney: See https://ocds-standard-development-handbook.readthedocs.io/en/latest/coding/index.html#id1

I upgraded pip-tools in a new commit.

kingfisher_scrapy/base_spider.py
jpmckinney: @yolile Can you answer this question? Otherwise, PR looks good.

kingfisher_scrapy/util.py
yolile: great! I thought that only the documentation about the standard development was there. Should we add a reference to that in kingfisher somewhere?

kingfisher_scrapy/base_spider.py
yolile: Sorry, I forgot about this one! the grouper method returns nulls values to complete the number of items set at 'size' parameter

kingfisher_scrapy/util.py
jpmckinney: I opened an issue to discuss what to do with that content: https://github.com/open-contracting/standard-development-handbook/issues/225

kingfisher_scrapy/base_spider.py
jpmckinney: Aha, thanks!

PR #370, opened 2020-04-27
docs: Add username and password to "Configure Kingfisher Scrape"

odscjames:
This section is linked directly from
https://ocdsdeploy.readthedocs.io/en/latest/use/kingfisher-collect.html#update-spiders-in-kingfisher-scrape
and that says nothing about usernames and passwords, so we should make sure someone trying to update spiders sees something to prompt them

### DIFF
--- a/docs/scrapyd.rst
+++ b/docs/scrapyd.rst
@@ -54,6 +54,15 @@ Create a ``~/.config/scrapy.cfg`` file using the template below, and set the ``u
    url = http://localhost:6800/
    project = kingfisher
 
+If you need a username and password to access the remote server, you can put them in here:
+
+.. code-block:: ini
+
+   [deploy:kingfisher]
+   ...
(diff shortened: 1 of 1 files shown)


odscjames:
I just fell afoul of this doing my first deploy directly from laptop after helping an analyst by fixing a bug in a spider, so good to note. (Note, setting a netrc file did not work for me for some reason)

jpmckinney:
What did your `.netrc` file contain and what were its user permissions?

jpmckinney:
You also need to use `-n` with `curl`. I made this clearer here: https://ocdsdeploy.readthedocs.io/en/latest/use/kingfisher-collect.html#create-a-netrc-file

Using `.netrc` is easier than setting the username and password in multiple locations, so I prefer to not put it in `scrapy.cfg`.

PR #379, opened 2020-05-12
Document and add from_date parameter to paraguay dncp spider

### LINKED ISSUE #378: Add date parameter to paraguay dncp's spider

yolile:
closes #378

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -41,10 +41,10 @@ class BaseSpider(scrapy.Spider):
 
     MAX_SAMPLE = 10
     MAX_RELEASES_PER_PACKAGE = 100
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -41,15 +41,18 @@ class BaseSpider(scrapy.Spider):
 
     MAX_SAMPLE = 10
     MAX_RELEASES_PER_PACKAGE = 100
+    VALID_DATE_FORMATS = {'year_month_day': '%Y-%m-%d', 'year_month_day_time': '%Y-%m-%dT%H:%M:%S'}
```
jpmckinney, line 44: Should we simply name these `date` and `datetime`? I don't think we need to support multiple date formats or multiple datetime formats (simpler for users).

Otherwise, looks good!

jpmckinney: I approved changes so that you can merge if you agree with the suggested change.

yolile: you are right, done!

PR #381, opened 2020-05-19
Add KingfisherStoreFiles extension to avoid storing files in spiders

### LINKED ISSUE #277: Use item_scraped signal in extension, instead of writing files in spider

yolile:
closes #277

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -86,47 +86,26 @@ class BaseSpider(scrapy.Spider):
 
         return spider
 
-    def get_local_file_path_including_filestore(self, filename):
+    def save_response_to_disk(self, response, filename, data_type=None, encoding='utf-8', post_to_api=True):
         """
-        Prepends Scrapy's storage directory and the crawl's relative directory to the filename.
+        Returns an item to yield, based on the response to a request.
         """
-        return os.path.join(self.crawler.settings['FILES_STORE'], self._get_crawl_path(), filename)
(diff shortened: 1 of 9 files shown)


jpmckinney:
@yolile To address the issue you identified, can you add to `settings.py`:

```python
LOG_LEVEL = os.getenv('SCRAPY_LOG_LEVEL', 'INFO')
```

https://docs.scrapy.org/en/latest/topics/settings.html#log-level

jpmckinney:
Actually, other DEBUG messages are useful (seeing which requests were retried, etc.). Let's instead add a custom log formatter, so that it excludes the OCDS data: https://docs.scrapy.org/en/latest/topics/logging.html#custom-log-formats

yolile:
@jpmckinney I added all your suggestions (changing georgia and digiwhist_base spiders to be able to move `get_local_file_path_including_filestore` etc methods inside the store extension, FYI the georgia one wasn't working now anyway)
And also added a `post_to_api` field to the item to no post to kingfisher-process zip and tar files.
Also, I don't store the item if the item has the `number` attribute as this attribute is used when a big JSON file is sent to kingfisher-process as the behavior was to store only the original data.

jpmckinney:
I did some refactoring and other cleanup. Looks good!

PR #385, opened 2020-05-27
Last release date per publisher command

### LINKED ISSUE #283: Make it easy to determine the latest release date across all sources

yolile:
closes #283


yolile:
@jpmckinney this is how the last release date feature is looking like, could you review and tell me if it is what you expected?

kingfisher_scrapy/settings.py
```diff
@@ -67,6 +67,7 @@
 #    'scrapy.extensions.telnet.TelnetConsole': None,
 #}
 EXTENSIONS = {
+   'kingfisher_scrapy.extensions.KingfisherLastDate': 0,
```
jpmckinney, line 70: 0 is falsy, so let's set it to 1, just in case 0 causes a bug.

kingfisher_scrapy/spiders/canada_montreal.py
```diff
@@ -10,14 +10,17 @@ class CanadaMontreal(BaseSpider):
     page_limit = 10000
 
     def start_requests(self):
+        if self.last:
+            self.page_limit = 1
```
jpmckinney, line 14: Won't CLOSESPIDER_ITEMCOUNT cause the spider to close before the 10000 page limit is reached?

kingfisher_scrapy/spiders/canada_buyandsell.py
```diff
@@ -9,27 +9,30 @@ class CanadaBuyAndSell(BaseSpider):
 
     def start_requests(self):
         yield scrapy.Request(
-            url='https://buyandsell.gc.ca/cds/public/ocds/tpsgc-pwgsc_ocds_EF-FY-13-14.json',
-            meta={'kf_filename': '13-14.json'}
+            url='https://buyandsell.gc.ca/cds/public/ocds/tpsgc-pwgsc_ocds_EF-FY-16-17.json',
+            meta={'kf_filename': '16-17.json'}
         )
-        if self.sample:
+        if self.sample or self.last:
```
jpmckinney, line 15: Won't CLOSESPIDER_ITEMCOUNT cause the spider to close before all requests are completed?

kingfisher_scrapy/extensions.py
```diff
@@ -6,10 +6,29 @@
...
 
 
+# https://docs.scrapy.org/en/latest/topics/extensions.html#writing-your-own-extension
+class KingfisherLastDate:
+    def __init__(self, filename):
+        self.filename = filename
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        filename = os.path.join(crawler.settings['LAST_RELEASE_DATE_FILE_PATH'], 'last_dates.txt')
```
jpmckinney, line 20: Let's prefix with `KINGFISHER_` to keep all our customs settings in one namespace.

kingfisher_scrapy/commands/last_release_date.py
```diff
@@ -0,0 +1,30 @@
...
+    def short_desc(self):
+        return 'Get the last published release date per publisher'
+
+    def run(self, args, opts):
+        settings = get_project_settings()
+        settings.set('CLOSESPIDER_ITEMCOUNT', 1)
+        settings.set('CONCURRENT_REQUESTS', 1)
+        process = CrawlerProcess(settings=settings)
+
+        process.crawl(AfghanistanRecords, last='true')
```
jpmckinney, line 23: You can instead loop over all spiders as done here: https://github.com/open-contracting/kingfisher-collect/blob/master/tests/test_spiders.py#L19-L22

jpmckinney:
In terms of the approach, updating all spiders to yield a different type of item seems like a lot of work. Could we instead add an [item pipeline](https://docs.scrapy.org/en/latest/topics/item-pipeline.html) step to take the File and FileItem items and return LastReleaseDate items? This would have a big if-elif to get the most recent date for each `data_type` (like `json.loads(response.text)[data_type][0]['date']` for the 'release_package' data type). In other words, we'd move the changes to the base spider into the item pipeline.

kingfisher_scrapy/commands/last_release_date.py
yolile: Yes, I thought that but I don't know if we want/can check the date for all the spiders, (for example, we have Afghanistan records and releases, we want to check both?) or there are spiders that doesn't support the sorting by date, so I don't know if it will be possible to have the last date information. But maybe we can add a parameter or annotation to the spider to indicate if the last date feature is available?

yolile:
> (If the first release in a package is not guaranteed to be the most recent, we can just do use min.)

do you mean max?

jpmckinney:
> > (If the first release in a package is not guaranteed to be the most recent, we can just do use min.)
> 
> do you mean max?

Yes, sorry, max.

kingfisher_scrapy/commands/last_release_date.py
jpmckinney: We can add a class attribute to 'opt-out' of the command, maybe something like `skip_last_release_date = True`, and the command can log which are skipped.

If there are many different reasons for skipping, we can maybe instead do `skip_last_release_date = "reason"`, and the command can log the reason as well.

yolile:
@jpmckinney I think the Item Pipeline is not supposed to be used to generate more Items, but just for modify/process the received one, the docs says:

> They receive an item and perform an action over it, also deciding if the item should continue through the pipeline or be dropped and no longer processed.

So maybe I can do what you suggested but directly on the Extension. So the extension checks if the "last" argument is set, if the item is a File or FileItem, the data type, and according to that extracts the date and saves it to the file. Or yield the LastReleaseDate item and the Pipeline save it to the disk. What do you think?

kingfisher_scrapy/spiders/canada_buyandsell.py
yolile: I tested it and besides CLOSESPIDER_ITEMCOUNT I set [CLOSESPIDER_PAGECOUNT](https://docs.scrapy.org/en/latest/topics/extensions.html#closespider-pagecount) to 0 too and now it works as expected

jpmckinney: If a spider needs to first download a file (that doesn't yield items) and then download additional files linked from that file, won't CLOSESPIDER_PAGECOUNT cause the spider to stop after that first file?

jpmckinney:
> @jpmckinney I think the Item Pipeline is not supposed to be used to generate more Items, but just for modify/process the received one, the docs says:
> 
> > They receive an item and perform an action over it, also deciding if the item should continue through the pipeline or be dropped and no longer processed.
> 
> So maybe I can do what you suggested but directly on the Extension. So the extension checks if the "last" argument is set, if the item is a File or FileItem, the data type, and according to that extracts the date and saves it to the file. Or yield the LastReleaseDate item and the Pipeline save it to the disk. What do you think?

The [process_item](https://docs.scrapy.org/en/latest/topics/item-pipeline.html#process_item) docs are:

> This method is called for every item pipeline component. process_item() must either: return a dict with data, return an Item (or any descendant class) object, return a Deferred or raise DropItem exception. Dropped items are no longer processed by further pipeline components.

My suggestion is that if it gets a File or FileItem as input, it can return a LastReleaseDate as output (if it receives FileError, we can either return a LastReleaseDate so that we still get output, but where the date is unknown – or we can raise DropItem and log the error). It's not generating new items, but converting existing items.

If we make the extension do the transformation instead, then we'll need another way to turn off the other extensions (currently, we can just make them ignore LastReleaseDate items).

kingfisher_scrapy/spiders/canada_buyandsell.py
yolile: Yes, you are right, it failed with australia_nsw for example. So maybe I will need to add a control to don't write duplicates entries in the extension.

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,39 @@
...
+        settings.set('CLOSESPIDER_ERRORCOUNT', 1)
+
+        process = CrawlerProcess(settings=settings)
+        spiders = process.spider_loader.list()
+        current_year = datetime.today().year
+        filename = get_skipped_output_file(settings)
+        for spider in spiders:
+            spider_cls = process.spider_loader.load(spider)
+            if hasattr(spider_cls, 'skip_latest_release_date'):
+                with open(filename, 'a+') as output:
```
jpmckinney, line 27: Instead of reopening the file, why not open it once in write mode before the loop?

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,39 @@
...
+                    output.write(f'Skipping {spider} because of {spider_cls.skip_latest_release_date} \n')
```
jpmckinney, line 28:
```suggestion
                    output.write(f'Skipping {spider}. Reason: {spider_cls.skip_latest_release_date}\n')
```

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,39 @@
...
+            else:
+                process.crawl(spider, latest='true', year=current_year)
+        process.start()
+
+
+def get_skipped_output_file(settings):
+    path = settings['KINGFISHER_LATEST_RELEASE_DATE_FILE_PATH']
+    if os.path.exists(path) and os.path.isdir(path):
+        shutil.rmtree(path)
```
jpmckinney, line 37: If we use write mode above, we can skip deleting the file here.

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,39 @@
...
+    os.makedirs(path)
```
jpmckinney, line 38:
```suggestion
    os.makedirs(path, exist_ok=True)
```

kingfisher_scrapy/spiders/honduras_oncae.py
```diff
@@ -14,6 +14,7 @@ class HondurasONCAE(ZipSpider):
     """
     name = 'honduras_oncae'
     data_type = 'release_package'
+    skip_latest_release_date = 'Already covered by honduras_portal_releases'
```
jpmckinney, line 17: For my information, what is the difference between oncae.gob.hn and www.contratacionesabiertas.gob.hn?

kingfisher_scrapy/spiders/dominican_republic.py
```diff
@@ -17,6 +17,7 @@ class DominicanRepublic(BaseSpider):
         Downloads a release package for the oldest year (2018, first link in the downloads page).
     """
     name = 'dominican_republic'
+    skip_latest_release_date = 'Rar file'
```
jpmckinney, line 20: Do we mean:

```suggestion
    skip_latest_release_date = "This command doesn't yet support identifying the latest release in a RAR file."
```

kingfisher_scrapy/extensions.py
```diff
@@ -7,10 +7,40 @@
...
+        path = crawler.settings['KINGFISHER_LATEST_RELEASE_DATE_FILE_PATH']
+        if not os.path.exists(path):
+            os.makedirs(path)
+        filename = os.path.join(path, 'latest_dates.txt')
+        extension = cls(filename=filename)
+        crawler.signals.connect(extension.item_scraped, signal=signals.item_scraped)
+        crawler.signals.connect(extension.spider_closed, signal=signals.spider_closed)
+        return extension
+
+    def item_scraped(self, item, spider):
```
jpmckinney, line 31: Let's add a `spider_opened` callback to truncate the file (i.e. just open it for writing).

kingfisher_scrapy/extensions.py
```diff
@@ -7,10 +7,40 @@
...
+class KingfisherLatestDate:
+    def __init__(self, filename):
+        self.filename = filename
+        self.written = {}
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        path = crawler.settings['KINGFISHER_LATEST_RELEASE_DATE_FILE_PATH']
+        if not os.path.exists(path):
+            os.makedirs(path)
```
jpmckinney, line 24:
```suggestion
        os.makedirs(path, exist_ok=True)
```

kingfisher_scrapy/extensions.py
```diff
@@ -7,10 +7,40 @@
...
+        crawler.signals.connect(extension.item_scraped, signal=signals.item_scraped)
+        crawler.signals.connect(extension.spider_closed, signal=signals.spider_closed)
+        return extension
+
+    def item_scraped(self, item, spider):
+        if not isinstance(item, LatestReleaseDateItem) or spider.name in self.written.keys():
+            return
+        self.written[spider.name] = True
+        with open(self.filename, 'a+') as output:
+            output.write(f"{spider.name}: {item['date']} \n")
```
jpmckinney, line 36: Let's use a comma so that it ends up being a CSV.

kingfisher_scrapy/base_spider.py
```diff
@@ -304,6 +306,7 @@ def start_requests(self):
 
     encoding = 'utf-8'
     zip_file_format = None
+    skip_latest_release_date = 'Zip file'
```
jpmckinney, line 309: Do we mean:

```suggestion
    skip_latest_release_date = "This command doesn't yet support identifying the latest release in a ZIP file."
```

kingfisher_scrapy/extensions.py
jpmckinney: Sorry, nevermind – I think the extension is initialized for each spider, which means spider_opened and engine_started will be called for each spider. So, we can just truncate in the command.

kingfisher_scrapy/spiders/honduras_oncae.py
yolile: Contrataciones Abiertas publishes joined data from ONCAE and the Ministry of Finance (they are the one who are using kingfisher for that)

kingfisher_scrapy/commands/latestreleasedate.py
yolile: I added your suggestion but still delete the directory as the "latest_dates.csv" file is also in that directory, and if I don't delete it we may end up having duplicates in the file, as the extension appends lines to it, and I cannot delete it or open it in write mode there as each spider open it each time.

kingfisher_scrapy/spiders/honduras_oncae.py
jpmckinney: Good to know. We can put this sort of information in the eventual data registry.

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,38 @@
...
+            for spider in spiders:
+                spider_cls = process.spider_loader.load(spider)
+                if hasattr(spider_cls, 'skip_latest_release_date'):
+                    output.write(f'Skipping {spider}. Reason: {spider_cls.skip_latest_release_date}\n')
+                else:
+                    process.crawl(spider, latest='true', year=current_year)
+        process.start()
+
+
+def get_skipped_output_file(settings):
```
jpmckinney, line 34: Let's move this method into the main method. It's a bit unusual for a function named `get_skipped_output_file` to do destructive things like deleting directories.

kingfisher_scrapy/commands/latestreleasedate.py
```diff
@@ -0,0 +1,38 @@
...
+    path = settings['KINGFISHER_LATEST_RELEASE_DATE_FILE_PATH']
+    shutil.rmtree(path, ignore_errors=True)
```
jpmckinney, line 36: Should we just `os.unlink` the CSV file?

kingfisher_scrapy/extensions.py
```diff
@@ -7,10 +7,39 @@
...
+        return extension
+
+    def item_scraped(self, item, spider):
+        if not isinstance(item, LatestReleaseDateItem) or spider.name in self.written.keys():
+            return
+        self.written[spider.name] = True
+        with open(self.filename, 'a+') as output:
+            output.write(f"{spider.name},{item['date']}\n")
+
+    def spider_closed(self, spider):
```
jpmckinney, line 37: We can add the `reason` argument, and then write that value instead of `error`.

https://docs.scrapy.org/en/latest/topics/signals.html#scrapy.signals.spider_closed

kingfisher_scrapy/extensions.py
```diff
@@ -7,10 +7,39 @@
...
-from kingfisher_scrapy.items import File, FileError, FileItem
+from kingfisher_scrapy.items import File, FileError, FileItem, LatestReleaseDateItem
 from kingfisher_scrapy.kingfisher_process import Client
 
 
+# https://docs.scrapy.org/en/latest/topics/extensions.html#writing-your-own-extension
+class KingfisherLatestDate:
+    def __init__(self, filename):
+        self.filename = filename
+        self.written = {}
```
jpmckinney, line 18: Let's instead use a set, and call it `spiders_seen`.

kingfisher_scrapy/extensions.py
```diff
@@ -118,9 +147,9 @@ def item_scraped(self, item, spider):
         """
         Sends an API request to store the file, file item or file error in Kingfisher Process.
         """
-        if not item.get('post_to_api', True):
-            return
 
+        if not item.get('post_to_api', True) or isinstance(item, LatestReleaseDateItem):
```
jpmckinney, line 151: Do we also need something in this extensions `spider_closed` callback? Or is it guaranteed to always be `processed` or otherwise not `finished`?

kingfisher_scrapy/settings.py
```diff
@@ -70,6 +70,7 @@
 #    'scrapy.extensions.telnet.TelnetConsole': None,
 #}
 EXTENSIONS = {
+    'kingfisher_scrapy.extensions.KingfisherLatestDate': 1,
```
jpmckinney, line 73: Let's put this after SentryLogging, so that it's in priority order for readers.

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
 
         return item
+
+
+class LatestReleaseDate:
+    def __init__(self):
+        self.processed = set()
+
+    def process_item(self, item, spider):
+        if spider.latest and (isinstance(item, FileItem) or isinstance(item, File)):
```
jpmckinney, line 50: We can use this syntax instead:

```suggestion
        if spider.latest and isinstance(item, (File, FileItem)):
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+            if spider.name in self.processed:
+                spider.crawler.engine.close_spider(self, reason='proccesed')
```
jpmckinney, line 52:
```suggestion
                spider.crawler.engine.close_spider(self, reason='processed')
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
         return item
+
+
+class LatestReleaseDate:
+    def __init__(self):
+        self.processed = set()
+
+    def process_item(self, item, spider):
+        if spider.latest and (isinstance(item, FileItem) or isinstance(item, File)):
+            if spider.name in self.processed:
```
jpmckinney, line 51: We can move this before the first if-statement.

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                spider.crawler.engine.close_spider(self, reason='proccesed')
+                return
+            date = None
+            data = json.loads(item['data'])
+            if item['data_type'] == 'release_package' or item['data_type'] == 'release' \
+                    or item['data_type'] == 'release_list' or item['data_type'] == 'compiled_release'\
+                    or item['data_type'] == 'release_package_list'\
+                    or item['data_type'] == 'release_package_list_in_results':
```
jpmckinney, line 59:
```suggestion
            if item['data_type'] in ('release_package', 'release_package_list', 'release_package_list_in_results',
                                     'release_list', 'release', 'compiled_release'):
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                if item['data_type'] == 'release_list':
+                    data = data
+                elif item['data_type'] == 'release_package':
```
jpmckinney, line 62:
```suggestion
                if item['data_type'] == 'release_package':
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                elif item['data_type'] == 'release_package_list_in_results':
+                    data = data['results'][0]['releases']
+                if data:
+                    if item['data_type'] == 'release' or item['data_type'] == 'compiled_release':
+                        date = data['date']
+                    else:
+                        date = max(r['date'] for r in data)
+            elif item['data_type'] == 'record_package' or item['data_type'] == 'record' or \
+                    item['data_type'] == 'record_list' or item['data_type'] == 'record_package_list' \
+                    or item['data_type'] == 'record_package_list_in_results':
```
jpmckinney, line 75:
```suggestion
            elif item['data_type'] in ('record_package', 'record_package_list', 'record_package_list_in_results',
                                       'record_list', 'record'):
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                if item['data_type'] == 'record_package':
+                    data = data['records']
+                elif item['data_type'] == 'record_package_list':
+                    data = data[0]['records']
+                elif item['data_type'] == 'record_package_list_in_results':
+                    data = data['results'][0]['records']
+                elif item['data_type'] == 'record_list':
+                    data = data
```
jpmckinney, line 83:
```suggestion
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                if item['data_type'] == 'release_list':
+                    data = data
+                elif item['data_type'] == 'release_package':
+                    data = data['releases']
+                elif item['data_type'] == 'release_package_list':
+                    data = data[0]['releases']
+                elif item['data_type'] == 'release_package_list_in_results':
+                    data = data['results'][0]['releases']
+                if data:
+                    if item['data_type'] == 'release' or item['data_type'] == 'compiled_release':
```
jpmckinney, line 69:
```suggestion
                    if item['data_type'] in ('release', 'compiled_release'):
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                elif item['data_type'] == 'record_package_list':
+                    data = data[0]['records']
+                elif item['data_type'] == 'record_package_list_in_results':
+                    data = data['results'][0]['records']
+                elif item['data_type'] == 'record_list':
+                    data = data
+                elif item['data_type'] == 'record':
+                    data = [data]
+                if data:
+                    data = data[0]
```
jpmckinney, line 87: Let's add a comment:

```
# This assumes that the first record in the record package has the most recent date.
```

kingfisher_scrapy/pipelines.py
```diff
@@ -40,3 +40,58 @@ def process_item(self, item, spider):
...
+                    if 'releases' in data:
+                        date = max(r['date'] for r in data['releases'])
+                    elif 'compiledRelease' in data:
+                        date = data['compiledRelease']['date']
+            self.processed.add(spider.name)
+            return LatestReleaseDateItem({
+                'date': date
+            })
```
jpmckinney, line 95:
```suggestion
            return LatestReleaseDateItem({'date': date})
```

kingfisher_scrapy/extensions.py
yolile: good point! I think most of the time the reason will be `closespider_itemcount` but I will add an extra if just in case (eg the API returns just 1 page etc)

kingfisher_scrapy/spiders/dominican_republic.py
```diff
@@ -13,7 +13,9 @@ class DominicanRepublic(CompressedFileSpider):
         Downloads a release package for the oldest year (2018, first link in the downloads page).
     """
     name = 'dominican_republic'
-    skip_latest_release_date = "This command doesn't yet support identifying the latest release in a RAR file."
```
jpmckinney: @yolile I think you meant to keep this? I'm making some commits, so I'll do it.

yolile:
No, because it is in its parent class now

yolile:
@jpmckinney all good, I added one more test to test `DropItem` for `FileError`. I'm not sure how to test the `spider.close` if

jpmckinney:
Thanks! Since this is a management command, there's less risk around untested code, so I think that's okay.

jpmckinney:
🎉 🎊 Do you want to send a CRM update to let people know we can now easily check MEL 1.5 (recent data)?

yolile:
Sure, I will!

PR #386, opened 2020-05-27
Add sentry logger as extension

### LINKED ISSUE #279: Send API errors to Sentry, instead of just logging them as warnings
I don't think anyone is checking the logs.

We can continue to log (maybe useful during scraper development, if the developer caused malformed requests to be created), but we should also send the errors to Sentry.

We'd need to add this project to Sentry. This might be a good time for OCP to create its own Team account: https://sentry.io/pricing/

yolile:
closes #279


kingfisher_scrapy/extensions.py
```diff
@@ -157,3 +158,20 @@ def _request(self, item, spider, method, *args, name='API'):
...
+    Send exceptions and errors to Sentry.
+    """
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        sentry_dsn = crawler.settings.get('SENTRY_DSN', None)
+        if sentry_dsn is None:
+            raise NotConfigured
+        extension = cls()
+        # by default only the errors are sent you sentry https://docs.sentry.io/platforms/python/logging/
```
jpmckinney, line 175: Let's put this in the docstring instead.
```suggestion
```

kingfisher_scrapy/extensions.py
```diff
@@ -157,3 +158,20 @@ def _request(self, item, spider, method, *args, name='API'):
         if not response.ok:
             spider.logger.warning(
                 'Failed to post [{}]. {} status code: {}'.format(item['url'], name, response.status_code))
+
+
+# https://stackoverflow.com/questions/25262765/handle-all-exception-in-scrapy-with-sentry
+class SentryLogging:
+    """
+    Send exceptions and errors to Sentry.
```
jpmckinney, line 166:
```suggestion
    Sends exceptions and log records to Sentry. Log records with a level of ``ERROR`` or higher are captured as events.

    https://docs.sentry.io/platforms/python/logging/
```

kingfisher_scrapy/settings.py
```diff
@@ -71,6 +71,8 @@
     # request is sent to Kingfisher Process.
     'kingfisher_scrapy.extensions.KingfisherFilesStore': 100,
     'kingfisher_scrapy.extensions.KingfisherProcessAPI': 500,
+    # before any other extension
+    'kingfisher_scrapy.extensions.SentryLogging': -1,
```
jpmckinney, line 75: Put the dict in priority order (i.e. with SentryLogging first). We can remove the comment, since the StackOverflow link explains the priority.

kingfisher_scrapy/settings.py
```diff
@@ -79,6 +81,8 @@
 #    'kingfisher_scrapy.pipelines.KingfisherScrapyPipeline': 300,
 #}
 
+# To send the logs to sentry
```
jpmckinney, line 84:
```suggestion
# To send exceptions and log records to Sentry.
```

kingfisher_scrapy/settings.py
```diff
@@ -79,6 +81,8 @@
...
+SENTRY_DSN = os.getenv('SENTRY_DSN')
 # To send items to Kingfishet Process, set this to, for example, "http://kingfisher.example.com" (no trailing slash).
```
jpmckinney, line 86: Noticed an unrelated typo:

```suggestion
# To send items to Kingfisher Process, set this to, for example, "http://kingfisher.example.com" (no trailing slash).
```

jpmckinney:
Nevermind the comment about tests – we assume Sentry just works in other repositories.

yolile:
@jpmckinney I merged this with master changes and now the review is required again

Issue #347, opened 2020-04-03
Add docstrings to spider classes


jpmckinney:
Include:

* [x] API (or bulk download) documentation URL
* [x] List of spider arguments (e.g. Colombia has sample, year and page)
  * Since the sample argument doesn't do that exact same thing for all spiders, we should document what it does for each spider (example below).

For example (this will render as definition lists in Sphinx):

```rst
API documentation
  https://www.colombiacompra.gov.co/transparencia/api
Swagger API documentation
  https://apiocds.colombiacompra.gov.co:8443/apiCCE2.0/
Spider arguments
  sample
    Download only the first page of results.
  page
    The page number from which to start crawling.
  year
    The year to crawl. See API documentation for valid values.
```

While we're at it, we should rename Colombia's `start` to `start_page`.

Once we start, we can add Sphinx's `automodule` for these docstrings to be rendered in the documentation.

jpmckinney:
@aguilerapy Thanks to your work in #350, we can now have nice docs like: https://kingfisher-scrape.readthedocs.io/en/latest/spiders.html#openopps, so that users can easily see what options each spider supports.

aguilerapy:
With #389 and #393 we cover all the spiders availables except the digiwhist ones, we can close this or there is something else to do?

jpmckinney:
I think that’s everything! Thanks!

PR #389, opened 2020-05-28
Add docstrings to spider classes 

aguilerapy:
I will create multiple PRs for #347. For a simpler review, I will try to have each reviewer verify the spiders he wrote.

In this first PR, most of the spiders were written by @yolile and me.


kingfisher_scrapy/spiders/nepal_dhangadhi.py
```diff
@@ -8,6 +8,11 @@
 
 
 class NepalDhangadhi(SimpleSpider):
+    """
+    Spider arguments
```
yolile, line 12: maybe we can add this https://ims.susasan.org/dhangadhi/about as "Bulk Download Documentation"

kingfisher_scrapy/spiders/nepal_portal.py
```diff
@@ -7,6 +7,11 @@
 
 
 class NepalPortal(SimpleSpider):
+    """
+    Spider arguments
```
yolile, line 11: maybe we can add http://ppip.gov.np/downloads as "Bulk Download Documentation"

kingfisher_scrapy/spiders/portugal.py
```diff
@@ -8,6 +8,13 @@
 
 
 class Portugal(ZipSpider):
+    """
+    API documentation
+        https://dados.gov.pt/pt/apidoc/
```
yolile, line 13: this page also have a swagger documentation at the bottom

kingfisher_scrapy/spiders/uruguay_historical.py
```diff
@@ -4,6 +4,13 @@
 
 
 class UruguayHistorical(ZipSpider):
+    """
+    API documentation
```
yolile, line 8: this is not an API so maybe we should call it "Bulk Download Documentation"

jpmckinney:
Woohoo, https://kingfisher-collect.readthedocs.io/en/latest/spiders.html is starting to look good.

PR #413, opened 2020-06-02
Add dryrun command and clean up issues it identified

jpmckinney:


jpmckinney:
Coverage decrease is due to a new untested command (but the command is for testing purposes, so testing it is not critical).

PR #417, opened 2020-06-17
Warn on duplicated file name and file items

### LINKED ISSUE #395: Warn about non-unique file_name (and file_name/number combo)
#394 adds a validation item pipeline. We can add an init method to the pipeline to initialize sets for the `file_name`s of `File` items and the `(file_name, number)` tuples of `FileItem` items.

Then, in the `process_item` method, we can check if those are already in the set. If so, we log a warning (since this will cause strange behavior in Kingfisher Process). If not, we add to the set.

yolile:
closes #395


yolile:
@jpmckinney thank you! it seems like you still have to approve the merge

PR #420, opened 2020-06-18
Update validation pipeline to use a json schema

### LINKED ISSUE #403: Use JSON Schema to validate items
Right now there's a simple `validate` method to check for required fields: https://github.com/open-contracting/kingfisher-collect/blob/master/kingfisher_scrapy/items.py

We can instead write a JSON Schema to validate the data. We'll need the `jsonschema` and `rfc3987` packages, and we can then do:

```python
from jsonschema import FormatChecker
from jsonschema.validators import Draft4Validator as validator

# Initialize the schema once, in the item pipeline's `__init__` method
with open('path/to/schema.json') as f:
    schema = json.load(f)

self.validator = validator(schema, format_checker=FormatChecker())

# Then, to validate:
validator.validate(item)
```

The JSON Schema can use:

* An enum for `data_type`
* `"format": "uri"` for `url`
* `"pattern": "^[^/]$"` for `file_name` (no path separators)
* `"minLength": 1` for required strings
* `"minimum": 1` for `number`
* appropriate `"type"` for all

yolile:
closes #403


yolile:
@jpmckinney the json schema check is failing due to camel case codelist and fields, but I think that we cannot avoid this, right?

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Schema for Kingfisher Collect File, FileItem and FileError",
+  "description": "Schema for Kingfisher Collect File, FileItem and FileError",
```
jpmckinney, line 4: I think we can remove all `title` and `description` fields. I've now made an exception in standard-maintenance-scripts for any files under `item_schema/`.

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
...
+            "release_package_list",
+            "record_package_list_in_results",
+            "release_package_list_in_results",
+            "release_package_json_lines",
+            "record_package_json_lines",
+            "release_package_in_ocdsReleasePackage_in_list_in_results",
+            "release_in_Release"
+          ],
+          "openCodelist": false,
+          "codelist": "dataType.csv",
```
jpmckinney, line 74: I think we can omit the codelist. The schema is just for validation – not for documentation.

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
...
+      "title": "Kingfisher Item",
+      "description": "A generic item with file_name and url to be extended by other items",
+      "type": "object",
+      "properties": {
+        "file_name": {
+          "title": "File Name",
+          "description": "File Name",
+          "type": "string",
+          "pattern": "^[^/]*$",
+          "minLength": 1
```
jpmckinney, line 27: If you change `*` to `+`, you can omit `minLength`.

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
...
+        },
+        "url": {
+          "title": "URL",
+          "description": "URL",
+          "type": "string",
+          "format": "uri",
+          "minLength": 1
```
jpmckinney, line 34: An empty URI is invalid, so `minLength` can be omitted.

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
...
+            "record_package_list_in_results",
+            "release_package_list_in_results",
+            "release_package_json_lines",
+            "record_package_json_lines",
+            "release_package_in_ocdsReleasePackage_in_list_in_results",
+            "release_in_Release"
+          ],
+          "openCodelist": false,
+          "codelist": "dataType.csv",
+          "minLength": 1
```
jpmckinney, line 75: `enum` guarantees a length, so this can be omitted.

kingfisher_scrapy/item_schema/item_schema.json
```diff
@@ -0,0 +1,176 @@
...
+        },
+        "encoding": {
+          "title": "Encoding",
+          "description": "Encoding",
+          "type": [
+            "string",
+            "null"
```
jpmckinney, line 82: Can we test all optional fields without `"null"`? I'm pretty sure unset fields are simply omitted.

yolile: Sure, actually I added most of the changes because the script was failing and complaining about things like this one (and also the use of title and description, codelist, etc)

yolile:
@jpmckinney I'm still having some json schema errors

jpmckinney:
@yolile No worries - I'll relax the validations today. (They are strict because so far we've only used JSON Schema for the standard.)

kingfisher_scrapy/item_schema/file.json
```diff
@@ -0,0 +1,28 @@
...
+  "title": "File",
+  "type": "object",
+  "properties": {
+    "post_to_api": {
+      "type": [
+        "boolean"
+      ]
+    },
+    "path": {
+      "description": "For the KingfisherProcessAPI extension to read the file.",
```
jpmckinney, line 17: I think it's okay to remove all title and description fields. If these need documentation, they should be added in the code instead.

kingfisher_scrapy/pipelines.py
```diff
@@ -1,12 +1,25 @@
...
+import jsonref as jsonref
+from jsonschema import FormatChecker
+from jsonschema.validators import Draft4Validator
 
 
 class Validate:
+    def __init__(self):
+        self.validators = {}
+        schema_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'item_schema')
+        for item in ['file', 'file_error', 'file_item']:
```
jpmckinney, line 15: I typically use tuples instead of lists for constants like this, but it's not important.

kingfisher_scrapy/pipelines.py
```diff
@@ -1,12 +1,25 @@
...
+            filename = os.path.join(schema_path, f'{item}.json')
+            with open(filename) as f:
+                schema = jsonref.load(f, base_uri=pathlib.Path(os.path.join(schema_path), 'item_schema').as_uri())
```
jpmckinney, line 18: Hmm, `'item_schema'` is outside the `os.path.join` call (which we don't need if it's only one argument), and in any case `schema_path` already contains `item_schema`. I'm not sure how this works.

Since we're re-using the `base_uri` parameter value, we can set `schema_path` to it above.

kingfisher_scrapy/pipelines.py
```diff
@@ -1,12 +1,25 @@
...
+            class_name = ''.join(word.title() for word in item.split('_'))
```
jpmckinney, line 19: To make the code simpler, we can rename the files to exactly match the class names.

kingfisher_scrapy/pipelines.py
yolile: @jpmckinney I changed it but for some reason jsonref fails if I don't put 'item_schema' again in base_uri I'm not sure why

jpmckinney: Aha, I had to add a trailing slash to the URI.

PR #421, opened 2020-06-19
Add Moldova Mtender scraper

### LINKED ISSUE #419: Add spider for new Moldova endpoint
https://public.mtender.gov.md/tenders
http://public.eprocurement.systems/ocds/tenders/

**Link to CRM issue**
https://crm.open-contracting.org/issues/6128

**Priority**
medium

aguilerapy:
closes #419

### DIFF
--- a/kingfisher_scrapy/spiders/moldova_mtender.py
+++ b/kingfisher_scrapy/spiders/moldova_mtender.py
@@ -1,0 +1,34 @@
+import json
+
+from kingfisher_scrapy.base_spider import SimpleSpider
+from kingfisher_scrapy.util import components, handle_http_error, join, parameters, replace_parameter
+
+
+class MoldovaMTender(SimpleSpider):
+    """
+    Spider arguments
+      sample
+        Download only one set of releases.
+    """
+    name = 'moldova_mtender'
+    data_type = 'release_package'
+    base_url = 'http://public.eprocurement.systems/ocds/tenders/'
+
+    def start_requests(self):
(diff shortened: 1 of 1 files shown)


yolile:
Hmm as this scraper is actually using the same URL as https://github.com/open-contracting/kingfisher-collect/blob/master/kingfisher_scrapy/spiders/moldova.py shouldn't we update the existing one instead of creating a new one? @duncandewhurst could you confirm that is correct? In addition, https://github.com/open-contracting/kingfisher-collect/blob/master/kingfisher_scrapy/spiders/moldova.py is not working anymore so maybe it is a good idea to update it with the new URL.

duncandewhurst:
Is it using the same url?

I thought the old scraper used `https://public.mtender.gov.md/tenders/<ocid>` and the new one uses `http://public.eprocurement.systems/ocds/tenders/<ocid>`

Comparing an example ocid in the [new](http://public.eprocurement.systems/ocds/tenders/ocds-b3wdp1-MD-1539840280133) and [old](https://public.mtender.gov.md/tenders/ocds-b3wdp1-MD-1539840280133) endpoints there are differences in the structure of the response and the extensions declared, so I think it's best to have a separate scraper for the new endpoint.

duncandewhurst:
Does https://kingfisher-collect.readthedocs.io/en/latest/spiders.html#moldova need updating?

jpmckinney:
Indeed! I ran `scrapy updatedocs` and committed. Should be updated on ReadTheDocs soon.

PR #425, opened 2020-06-23
Add LinksSpiderError and update tests

### LINKED ISSUE #412: Change LinksSpider to error if the next link isn't found on the first page
It's okay for the next link to not be found on subsequent pages, but it should be required on the first page.

aguilerapy:
closes #412

### DIFF
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -12,3 +12,7 @@ class SpiderArgumentError(KingfisherScrapyError):
 
 class MissingRequiredFieldError(KingfisherScrapyError, KeyError):
     """Raised when an item is missing a required field"""
+
+
+class LinksSpiderError(KingfisherScrapyError):
+    """Raised when a next link is invalid in LinksSpider"""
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/exceptions.py
```diff
@@ -12,3 +12,7 @@ class SpiderArgumentError(KingfisherScrapyError):
 
 class MissingRequiredFieldError(KingfisherScrapyError, KeyError):
     """Raised when an item is missing a required field"""
+
+
+class LinksSpiderError(KingfisherScrapyError):
+    """Raised when a next link is invalid in LinksSpider"""
```
jpmckinney, line 18: Let's use a more descriptive class name, since it's possible the LinksSpider will implement other exceptions in the future:

```suggestion
class MissingNextLinkError(KingfisherScrapyError):
    """Raised when a next link is not found on the first page of results"""
```

tests/test_links_spider.py
```diff
@@ -65,7 +65,13 @@ def test_parse_200():
 def test_next_link_not_found():
     spider = spider_with_crawler(spider_class=LinksSpider)
     spider.next_page_formatter = lambda url: 'next.json'
+    body = '{"links": {"next": ""}}'
 
     with pytest.raises(KingfisherScrapyError) as e:
-        assert spider.next_link(response_fixture(body='{"links": {"next": ""}}'))
+        meta = {'file_name': 'test', 'depth': 0}
+        assert spider.next_link(response_fixture(meta=meta, body=body))
```
jpmckinney, line 72: We can remove the `assert` here.

PR #427, opened 2020-06-24
Update Moldova mtender scraper

### LINKED ISSUE #419: Add spider for new Moldova endpoint

aguilerapy:
closes #419

### DIFF
--- a/kingfisher_scrapy/spiders/moldova_mtender.py
+++ b/kingfisher_scrapy/spiders/moldova_mtender.py
@@ -1,7 +1,9 @@
 import json
 
+import scrapy
+
 from kingfisher_scrapy.base_spider import SimpleSpider
-from kingfisher_scrapy.util import components, handle_http_error, join, parameters, replace_parameter
+from kingfisher_scrapy.util import components, handle_http_error, parameters, replace_parameter
 
 
 class MoldovaMTender(SimpleSpider):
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/moldova_mtender.py
```diff
@@ -12,20 +12,20 @@ class MoldovaMTender(SimpleSpider):
     """
     name = 'moldova_mtender'
     data_type = 'release_package'
-    base_url = 'http://public.eprocurement.systems/ocds/tenders/'
 
     def start_requests(self):
         url = 'https://public.mtender.gov.md/tenders/'
         yield self.build_request(url, formatter=components(-1), callback=self.parse_list)
```
jpmckinney, line 18: Since `tenders` doesn't disambiguate requests, we can remove it.

```suggestion
        yield scrapy.Request(url, meta={'file_name': 'offset-0.json'}, callback=self.parse_list)
```

jpmckinney: I can't comment on the other line since it's not in the PR, but it would change:

```python
        yield self.build_request(url, formatter=join(components(-1), parameters('offset')), callback=self.parse_list)
```

to:

```python
        yield self.build_request(url, formatter=parameters('offset'), callback=self.parse_list)
```

Then we can remove the `join` import.

PR #431, opened 2020-06-25
Fix colombia

### LINKED ISSUE #430: colombia: Fix scraper file name
The filename is taken from the "page" parameter but this page remains the same through all the pages after the 2nd one. We should use the "id" parameter instead.

yolile:
closes #430

### DIFF
--- a/kingfisher_scrapy/item_schema/item.json
+++ b/kingfisher_scrapy/item_schema/item.json
@@ -49,8 +49,7 @@
         "encoding": {
           "type": "string"
         },
-        "data": {
-        }
+        "data": {}
       },
       "required": [
         "data",
--- a/kingfisher_scrapy/item_schema/item.json
+++ b/kingfisher_scrapy/item_schema/item.json
@@ -50,8 +50,6 @@
           "type": "string"
         },
         "data": {
-          "type": "string",
-          "minLength": 1
         }
       },
       "required": [
--- a/kingfisher_scrapy/item_schema/item.json
+++ b/kingfisher_scrapy/item_schema/item.json
@@ -49,7 +49,9 @@
         "encoding": {
           "type": "string"
         },
-        "data": {}
+        "data": {
+          "minLength": 1
+        }
       },
       "required": [
         "data",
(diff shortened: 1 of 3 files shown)


yolile:
@jpmckinney I also removed the "string" type from the "data" parameter at item Shema and left it empty as there isn't a type that we can use (object type didn't worked either)

jpmckinney:
Right – because it's binary sometimes?

kingfisher_scrapy/item_schema/item.json
```diff
@@ -49,10 +49,7 @@
         "encoding": {
           "type": "string"
         },
-        "data": {
-          "type": "string",
-          "minLength": 1
-        }
+        "data": {}
```
jpmckinney, line 52: Can we not keep "minLength": 1? I think it only runs when the type is a string.

Let's also add a test with both binary and string values for `data`.

kingfisher_scrapy/spiders/colombia.py
```diff
@@ -29,14 +29,21 @@ class Colombia(LinksSpider):
         If ``from_date`` is provided and ``until_date`` don't, defaults to today.
     """
     name = 'colombia'
-    next_page_formatter = staticmethod(parameters('page'))
+    next_page_formatter = staticmethod(parameters('_id'))
     default_from_date = '2011-01-01'
 
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super(Colombia, cls).from_crawler(crawler, date_format='date', *args, **kwargs)
```
jpmckinney, line 37: Can just do `super()`, no? (We might want to clean this up globally in a new issue.)

`date_format` is 'date' by default.

```suggestion
        spider = super().from_crawler(crawler, *args, **kwargs)
```

kingfisher_scrapy/spiders/colombia.py
```diff
@@ -29,14 +29,21 @@ class Colombia(LinksSpider):
...
+        if (spider.from_date or spider.until_date) and hasattr(spider, 'year'):
+            raise scrapy.exceptions.CloseSpider('The use of from and/or until with year parameter is not supported')
```
jpmckinney, line 39: Make the names of spider arguments more explicit:

```suggestion
            raise scrapy.exceptions.CloseSpider('You cannot specify both a year spider argument and from_date/until_date spider argument(s).')
```

PR #433, opened 2020-06-26
Use pkgutil.get_data to get json schema files

yolile:


yolile:
@jpmckinney great, I don't know why the coverage decreased, should I merge the changes anyway?

jpmckinney:
Yes, go ahead. Change is very small; it's likely because we reduced the number of lines in a file with good coverage.

Issue #408, opened 2020-06-01
chile_base: Add comment to explain status check


jpmckinney:
Does the Chile API return error objects with HTTP 200 status codes? If that's the case, let's add a comment, to explain why we check for `status`.

https://github.com/open-contracting/kingfisher-collect/pull/222

yolile:
done in #437

PR #437, opened 2020-06-29
Fix Chile Compra Bulk

### LINKED ISSUE #375: chile_compra_bulk: Spider doesn't recognize HTTP 500 errors
For example, responses are like:

```json
{
  "status": 500,
  "detail": "Error en la generación. Para más información dirijase a la ficha del documento. http://www.mercadopublico.cl/Procurement/Modules/RFB/DetailsAcquisition.aspx?idlicitacion="
}
```

While fixing this, let's also add documentation for the spider as described in #347, so that it's easier to discover API documentation, etc.

cc @Camilamila

Sentry Issue: [OPEN-CONTRACTING-KINGFISHERLIV-64](https://sentry.io/organizations/open-contracting-partnership/issues/1635755590/?referrer=github_integration)

```
Exception: Record list not found
(7 additional frame(s) were not displayed)
...
  File "ocdskingfisherprocess/web/views_api_v1.py", line 121, in dispatch_request
    request.form.get('local_file_name')
  File "ocdskingfisherprocess/store.py", line 96, in store_file_from_local
    self.store_file_from_data(filename, url, data_type, data, file_warnings=file_to_store.get_warnings())
  File "ocdskingfisherprocess/store.py", line 122, in store_file_from_data
    raise e
  File "ocdskingfisherprocess/store.py", line 118, in store_file_from_data
    self.store_file_item(filename, url, data_type, item_data, number)
  File "ocdskingfisherprocess/store.py", line 173, in store_file_item
    raise Exception("Record list not found")
```

yolile:
closes #375 #408


kingfisher_scrapy/spiders/chile_base.py
```diff
@@ -42,6 +42,11 @@ def start_requests(self):
     @handle_http_error
     def parse_list(self, response):
         data = json.loads(response.text)
+        # some files contain invalid packages, eg:
+        # {
+        #   "status": 500,
+        #   "detail": "error"
+        # }
         if 'status' in data and data['status'] != 200:
             yield self.build_file_error_from_response(response, errors={'http_code': data['status']})
```
jpmckinney, line 51: We can set `errors=data` to preserve the full error message body. To have a standard field for HTTP code, we can do `data['http_code'] = data['status']` first.

kingfisher_scrapy/spiders/chile_compra_bulk.py
```diff
@@ -27,6 +29,20 @@ def start_requests(self):
...
+        # some files contain invalid record packages, eg:
+        # {
+        #   "status": 500,
+        #   "detail": "error"
+        # }
+        if 'status' in json_data and json_data['status'] != 200:
+            return FileError({
+                'url': url,
+                'errors': {'http_code': json_data['status']},
+            })
```
jpmckinney, line 46: As with the other comment, we can set `'errors': json_data` to preserve the full error message body.

kingfisher_scrapy/spiders/chile_compra_bulk.py
```diff
@@ -27,6 +29,21 @@ def start_requests(self):
...
+        #   "detail": "error"
+        # }
+        if 'status' in json_data and json_data['status'] != 200:
+            json_data['http_code'] = json_data['status']
+            return FileError({
+                'url': url,
+                'errors': json_data,
+            })
+        else:
+            return super().build_file(data=data, file_name=file_name, url=url, data_type=data_type, encoding=encoding)
```
jpmckinney, line 49: Should we instead do `data=None, **kwargs` in the method signature, and then `super().build_file(data=data, **kwargs)` here?

kingfisher_scrapy/spiders/chile_base.py
```diff
@@ -67,5 +73,6 @@ def parse_list(self, response):
                 meta={
                     'year': year,
                     'month': month,
-                }
+                },
+                callback=self.parse_list
```
yolile, line 77: @jpmckinney sorry, I noticed that chile_base was missing a callback call in the parse_list method so I included the fix here too

kingfisher_scrapy/spiders/chile_compra_bulk.py
yolile: I think it is kind of the same, but I changed it

jpmckinney: Indeed – but this way we don't need to change two places if we add keyword arguments :)

PR #440, opened 2020-07-09
Ecuador spider

### LINKED ISSUE #418: Add spider for Ecuador COVID-19 emergency publication
https://portal.compraspublicas.gob.ec/sercop/data-estandar-ocds/

**Link to CRM issue**
https://crm.open-contracting.org/issues/4397

**Priority**
Low

yolile:
closes #418

### DIFF
--- a/kingfisher_scrapy/spiders/ecuador_emergency.py
+++ b/kingfisher_scrapy/spiders/ecuador_emergency.py
@@ -1,0 +1,42 @@
+import scrapy
+
+from kingfisher_scrapy.base_spider import SimpleSpider
+from kingfisher_scrapy.util import handle_http_error, components
+
+
+class EcuadorEmergency(SimpleSpider):
+    """
+    Bulk download documentation
+      https://portal.compraspublicas.gob.ec/sercop/data-estandar-ocds/
+    Spider arguments
+      sample
+        Downloads a release package for the oldest year (first link in the downloads page).
+    """
+    name = 'ecuador_emergency'
+    data_type = 'release_package'
+    custom_settings = {
+        'CONCURRENT_REQUESTS': 1,
+        'COOKIES_DEBUG': True
+    }
+
+    def start_requests(self):
+        yield scrapy.Request(
+            'https://portal.compraspublicas.gob.ec/sercop/data-estandar-ocds/',
+            meta={'file_name': 'list.html'},
+            callback=self.parse_list,
+        )
+
+    @handle_http_error
+    def parse_list(self, response):
+        for row in response.xpath('//tr'):
+            filename = row.xpath('td/p/strong/text()').extract_first()
+            base_link = row.xpath('td/strong/a/@href').extract_first()
+            if base_link:
+                filename = 'ocds-' + filename + '.json'
+                url = f'{base_link.replace("sharing", "fsdownload")}/{filename}'
+                yield self.build_request(base_link, formatter=components(-1),
+                                         meta={'next': url}, callback=self.with_cookie)
+
+    def with_cookie(self, response):
+        yield self.build_request(response.meta['next'],
+                                 formatter=components(-1))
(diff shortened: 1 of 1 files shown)


yolile:
@jpmckinney it is working now with this approach, one URL is not working now https://cloud.sercop.gob.ec:5001/sharing/OEZLv0TMB for that I disallowed the redirection and let the handle error in parse check the 302 status in a request with the correct cookie

kingfisher_scrapy/spiders/ecuador_emergency.py
```diff
@@ -0,0 +1,58 @@
...
+    urls = []
+
+    def start_requests(self):
+        yield scrapy.Request(
+            'https://portal.compraspublicas.gob.ec/sercop/data-estandar-ocds/',
+            meta={'file_name': 'list.html'},
+            callback=self.parse_list)
+
+    @handle_http_error
+    def parse_list(self, response):
```
jpmckinney, line 29: I find the choice of variable and dict-key names makes the code a little harder to read. What if we used `html_url` instead of `base_link` and `'base'` and `data_url` instead of `url` and `'data'`?

kingfisher_scrapy/spiders/ecuador_emergency.py
```diff
@@ -0,0 +1,58 @@
...
+        for row in response.xpath('//tr'):
+            filename = row.xpath('td/p/strong/text()').extract_first()
+            base_link = row.xpath('td/strong/a/@href').extract_first()
+            if base_link:
+                filename = 'ocds-' + filename + '.json'
+                url = f'{base_link.replace("sharing", "fsdownload")}/{filename}'
```
jpmckinney, line 35:
```suggestion
                url = f'{base_link.replace("sharing", "fsdownload")}/ocds-{filename}.json'
```

kingfisher_scrapy/spiders/ecuador_emergency.py
```diff
@@ -0,0 +1,58 @@
...
+                self.urls.append({'base': base_link, 'data': url})
+                if self.sample:
+                    break
+        # request the same url again just for go request_cookie_save_data method
+        yield self.build_request(response.request.url, callback=self.request_cookie_save_data,
+                                 formatter=components(-1),
+                                 dont_filter=True)
```
jpmckinney, line 42: Why do we need to request the same URL again, instead of calling `request_cookie_save_data()` directly?

kingfisher_scrapy/spiders/ecuador_emergency.py
```diff
@@ -0,0 +1,58 @@
...
+    def request_cookie_save_data(self, response):
+        if self.urls:
+            # we process one url and its response with its cookie at a time
+            url = self.urls.pop()
+            yield self.build_request(url['base'], meta={'next': url['data']},
+                                     formatter=components(-1), callback=self.with_cookie)
+        if 'data' in response.meta:
+            yield from self.parse(response)
+
+    def with_cookie(self, response):
```
jpmckinney, line 53: Should we rename this `parse_page` since it's the callback for individual pages?

kingfisher_scrapy/spiders/ecuador_emergency.py
```diff
@@ -0,0 +1,58 @@
...
+                                 dont_filter=True)
+
+    def request_cookie_save_data(self, response):
+        if self.urls:
+            # we process one url and its response with its cookie at a time
+            url = self.urls.pop()
+            yield self.build_request(url['base'], meta={'next': url['data']},
+                                     formatter=components(-1), callback=self.with_cookie)
+        if 'data' in response.meta:
+            yield from self.parse(response)
```
jpmckinney, line 51: I find it a bit confusing to have the same callback for data URLs and for the main page. If we move the `if self.urls` statement into its own method, we can call it from `parse_list` and then also from this callback (which we might rename as `parse_data`).

Issue #86, opened 2019-01-31
Check Armenia scraper


jpmckinney:
The old scraper has not been tested according to open-contracting/kingfisher#100

aguilerapy:
Running the Armenia scraper, the last page saved is https://armeps.am/ocds/release?limit=100&offset=1519411907725 , then, in the [next request page](https://armeps.am/ocds/release?limit=100&offset=1519716611092) return error 500 and finish the spider.

jpmckinney:
I think the action here is to document the limitations of the API (e.g. give an indication of how many pages of results get downloaded before the 500 error occurs).

aguilerapy:
So I will write the scraper and show the message of how much was downloaded when reaching the error 505, right? @jpmckinney

jpmckinney:
The spider runs – it just errors at some point (and we have no way to skip the bad page).

So, the action is to describe in the docstring about how many pages/results can be downloaded before the error occurs, e.g.

> The API consistently returns an irrecoverable error at ``offset=1519828326255``, after downloading about 120 pages (as of 2020-06-24).

jpmckinney:
Actually, that gave me an idea to get around the bug: we can just increase the timestamp until it starts working again. The lowest number I got was https://armeps.am/ocds/release?limit=100&offset=1519840056264

aguilerapy:
How does that manual binary search work? 

There could be another available `offset` between [1519828326255](https://armeps.am/ocds/release?limit=100&offset=1519411907725 ) and [1519840056264](https://armeps.am/ocds/release?limit=100&offset=1519840056264), right?

jpmckinney:
I took the last number that worked (1519411907725) as the minimum and and picked a number in the future as the maximum, then took the midpoint between them. If it worked, I set the midpoint to the maximum; otherwise, I set it to the minimum. Then did it all again.

It's possible that there are some numbers in between that work, but this is unlikely – and the only way to check is to advance the offset some small increment at a time, which will require a lot of failing requests.

aguilerapy:
How do you picked that number in the future as the maximum? Because that number directly influences what offsets we reach. For example if the maximum picked is `a`, the midpoint reached is `x`, on the other hand, if the maximum picked is `b`, the midpoint is `y`.

If the midpoint stops moving and the current `offset` doesn't work, do we continue from the last point that worked?

And what happens if we don't find any `offset` that works during the search? @jpmckinney

[...]

PR #442, opened 2020-07-10
Update Armenia scraper

aguilerapy:


kingfisher_scrapy/spiders/armenia.py
```diff
@@ -14,7 +19,94 @@ class Armenia(LinksSpider):
...
+            # continue the search with http error
+            return self.binary_search(response, succeed=False)
+
+        # save the page with next_page error
+        self.next_page_error = self.last_page_success
+        # start a binary search
+        return self.binary_search(response, start='start')
+
+    def binary_search(self, response, succeed=None, start=None):
+        microseconds_per_day = 86400000
```
yolile, line 59: @aguilerapy can you document a bit more about the response and how the binary search you are doing works?

aguilerapy:
Follow #86.

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -14,7 +19,110 @@ class Armenia(LinksSpider):
...
+        'CONCURRENT_REQUESTS': 1,
+    }
 
     def start_requests(self):
         url = 'https://armeps.am/ocds/release'
         yield scrapy.Request(url, meta={'file_name': 'offset-0.json'})
+
+    def parse_next_link(self, response):
+        # check if there is a search in process
+        binary_search = True if 'minimum' in response.request.meta else False
```
yolile, line 37:
```suggestion
        binary_search = 'minimum' in response.request.meta
```

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -14,7 +19,110 @@ class Armenia(LinksSpider):
...
+
+        if self.is_http_success(response):
+            # save this page for the lowest offset that works or the page with next_page error
+            self.last_page_success = response.request.url
```
yolile, line 41: maybe a clearer name for this variable could be `last_successful_page`

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -1,11 +1,16 @@
...
 import scrapy
 
 from kingfisher_scrapy.base_spider import LinksSpider
-from kingfisher_scrapy.util import parameters
+from kingfisher_scrapy.util import parameters, replace_parameter
 
 
 class Armenia(LinksSpider):
     """
+    If the API returns an error in ``next_page``, a binary search is performed to retrieve the success download cycle.
```
romifz, line 12:
```suggestion
    If the API returns an error in ``next_page``, a binary search is performed in an attempt to find the next working URL.
```

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -1,11 +1,16 @@
...
+    The search range extends for a maximum of 10 consecutive days from the last day of the successful download.
```
romifz, line 13:
```suggestion
    The search range extends for a maximum of 10 consecutive days from the timestamp in the ``offset``  parameter of the last successful URL.
```

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -14,7 +19,110 @@ class Armenia(LinksSpider):
...
+        if binary_search:
+            # continue the search with http error
+            return self.binary_search(response, succeed=False)
+
+        # save the page with next_page error
+        self.next_page_error = self.last_page_success
+        # start a binary search
+        return self.binary_search(response, start='start')
+
+    def binary_search(self, response, succeed=None, start=None):
```
romifz, line 58: I would change the name of the method here to `search_next_working_link` or something similar. Note that you are not only performing a binary search here, you are also moving forward and restarting the search if needed.

kingfisher_scrapy/spiders/armenia.py
aguilerapy: Done.

kingfisher_scrapy/base_spider.py
```diff
@@ -345,6 +345,7 @@ class LinksSpider(SimpleSpider):
        :meth:`~kingfisher_scrapy.base_spider.BaseSpider.build_request`
     1. Optionally, set a ``next_pointer`` class attribute to the JSON Pointer for the next link (default "/links/next")
     1. Write a ``start_requests`` method to request the first page of API results
```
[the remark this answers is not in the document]
aguilerapy: Done.

kingfisher_scrapy/spiders/armenia.py
aguilerapy: Done.

aguilerapy:
I updated the requested changes but I think there is a coverage issue outside of them.

kingfisher_scrapy/spiders/armenia.py
```diff
@@ -14,7 +21,117 @@ class Armenia(LinksSpider):
...
+        microseconds_per_day = 86400000
+        if start:
+            if start == 'start':
+                # start with last number that worked as the minimum
+                minimum = self.get_offset(self.next_page_error)
+            else:
+                # set the minimum to the maximum
+                minimum = self.get_offset(response.request.url)
+            # start with one day in the future of the current minimum as the maximum
+            maximum = minimum + (microseconds_per_day * self.days)
```
jpmckinney, line 92: My understanding is that this advances the maximum not by one day (as documented) but by `self.days`, since `minimum` is advancing, too.

jpmckinney:
I'm doing a review. The code is quite hard to read because it uses the same callback for two separate code paths, and not all requests opt-in to that callback, so it's very confusing. I think I'm almost done with a cleaner version.

aguilerapy:
OK! I was also cleaning the self.days that you told me.

jpmckinney:
I'm testing my changes now, but the main changes are:

* Use distinct callbacks for the three actions:
    1. Crawl normally
    1. Decide which date range on which to do a binary search
    1. Perform a binary search
* Set `dont_retry` in the request meta. The middleware is configured before `custom_settings` is read, so the `RETRY_TIMES` setting was having no effect in my testing.
* Since there are a lot of milliseconds in a day, it always takes 27 requests to finish a binary search (`86400000 > 2**26`). I tested with a few examples, and it seems enough to be within 45mins of the correct answer (5 requests).
  * Note: We can't simply check if two responses are identical, because Armenia's API sets the publishedDate and release date to be the current time... (We can still check with substitution, but it's more of a pain.)

jpmckinney:
Starting at 1521734640143, there is no good offset within 10 days. However, there is one within 64 days. I implemented exponential search with a limit of 10 requests (leads to the same number of requests as for 10 days).

PR #438, opened 2020-06-30
Add Scotland Proactis scraper

aguilerapy:
closes #426

PR #445, opened 2020-07-13
Add from_date and until_date parameters to uruguay_base

### LINKED ISSUE #444: Uruguay: add from_date and until_date parameters

yolile:
closes #444

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -42,7 +42,7 @@ class BaseSpider(scrapy.Spider):
 
     MAX_SAMPLE = 10
     MAX_RELEASES_PER_PACKAGE = 100
-    VALID_DATE_FORMATS = {'date': '%Y-%m-%d', 'datetime': '%Y-%m-%dT%H:%M:%S', 'month-year': '%m-%Y'}
+    VALID_DATE_FORMATS = {'date': '%Y-%m-%d', 'datetime': '%Y-%m-%dT%H:%M:%S', 'year-month': '%Y-%m'}
 
     def __init__(self, sample=None, note=None, from_date=None, until_date=None,
                  date_format='date', *args, **kwargs):
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -42,7 +42,7 @@ class BaseSpider(scrapy.Spider):
 
     MAX_SAMPLE = 10
     MAX_RELEASES_PER_PACKAGE = 100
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -42,7 +42,7 @@ class BaseSpider(scrapy.Spider):
 
     MAX_SAMPLE = 10
     MAX_RELEASES_PER_PACKAGE = 100
-    VALID_DATE_FORMATS = {'date': '%Y-%m-%d', 'datetime': '%Y-%m-%dT%H:%M:%S'}
+    VALID_DATE_FORMATS = {'date': '%Y-%m-%d', 'datetime': '%Y-%m-%dT%H:%M:%S', 'month-year': '%m-%Y'}
```
jpmckinney, line 45: YYYY-MM is preferred to conform to ISO 8601.

kingfisher_scrapy/spiders/uruguay_base.py
```diff
@@ -1,17 +1,21 @@
...
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        return super(UruguayBase, cls).from_crawler(crawler, date_format='month-year', *args, **kwargs)
 
     def start_requests(self):
         url = 'http://comprasestatales.gub.uy/ocds/rss/{0.year:d}/{0.month:02d}'
-
-        start = date(2017, 11, 1)
-        stop = date.today().replace(day=1)
+        start = self.from_date if self.from_date else datetime.strptime(self.default_from_date, self.date_format)
```
jpmckinney, line 17: Isn't the default value already parsed in base spider?

kingfisher_scrapy/spiders/uruguay_base.py
```diff
@@ -1,17 +1,21 @@
...
+        stop = self.until_date if self.until_date else date.today().replace(day=1)
```
jpmckinney, line 18: Won't the day component always be 1 when the date format is year-month, based on how until_date is evaluated in base spider?

kingfisher_scrapy/spiders/uruguay_base.py
yolile: only if the from_date or until_date parameters are set, otherwise don't.

kingfisher_scrapy/base_spider.py
yolile: agree, this was to conform with what @aguilerapy is doing in #438 but I will change it here

kingfisher_scrapy/spiders/uruguay_base.py
yolile: same as the other comment, if none of the parameters are set then the default values are not set either

PR #447, opened 2020-07-13
Add until_date to paraguay_dncp spider

### LINKED ISSUE #446: Paraguay DNCP: add until_date parameter

yolile:
closes #446


jpmckinney:
> Why is from_date passed in the meta? It doesn't look like we use it.

Once this question is answered / code is changed, I'm happy to approve.

yolile:
I'm not sure why that meta was used, but I removed it now as I didn't find a use for it.

PR #462, opened 2020-07-23
Add crawl_time and keep_collection_open spider parameters

yolile:


kingfisher_scrapy/extensions.py
```diff
@@ -154,9 +154,13 @@ def item_scraped(self, item, spider):
 
         if not item.get('post_to_api', True) or isinstance(item, LatestReleaseDateItem):
             return
+        date_format = '%Y-%m-%d %H:%M:%S'
+        collection_data_version = spider.get_start_time(date_format)
+        if spider.custom_collection_data_version:
+            collection_data_version = spider.custom_collection_data_version.strftime(date_format)
```
jpmckinney, line 160: Let's put the `collection_data_version = spider.get_start_time(date_format)` in an `else` branch.

kingfisher_scrapy/extensions.py
```diff
@@ -134,7 +134,7 @@ def spider_closed(self, spider, reason):
         Sends an API request to end the collection's store step.
         """
         # https://docs.scrapy.org/en/latest/topics/signals.html#spider-closed
-        if reason != 'finished' or spider.latest:
+        if reason != 'finished' or spider.latest or spider.custom_collection_data_version:
```
jpmckinney, line 137: Let's have a separate spider argument for leaving the collection open, to not overload the semantics of collection_data_version.

jpmckinney: I think we also need to change the call to get_start_time in this method and in the KingfisherFilesStore.item_scraped method.

We might want to move the preparation of the collection_source, collection_data_version, collection_sample dict to a common private method.

kingfisher_scrapy/extensions.py
jpmckinney: We might want to move the `strftime` call out of the `get_start_time` method, so that we can just call it once here (to do a small reduction in code duplication).

kingfisher_scrapy/base_spider.py
```diff
@@ -40,6 +40,12 @@ class BaseSpider(scrapy.Spider):
     .. code:: bash
 
         scrapy crawl spider_name -a note='Started by NAME.'
+
+    Use the same collection_id in Kingfisher Process for a source_id:
+
+     .. code:: bash
+
+        scrapy crawl spider_name -a custom_collection_data_version=2020-01-01
```
jpmckinney, line 48: I'm wondering about the name `custom_collection_data_version`.

From Kingfisher Collect's perspective, what we're doing is overriding the start_time used in the KingfisherFilesStore and KingfisherProcessAPI extensions. So maybe rename to `start_time`? That way, we're not letting Kingfisher Process' logic bleed in too much.

yolile: I wasn't sure about `custom_collection_data_version` either, although I don't want to confuse the users, (eg `start_date` sounds similar to `from_date`  etc), so maybe `collection_stat_time` ? or do you think that with the argument description will be enough?

jpmckinney: Good point. Maybe `crawl_time`?

From Kingfisher Collect's perspective, we can explain: Each crawl writes data to its own directory. By default, this directory is named according to the time the crawl started. To override the time (for example, to force a new crawl to write to the same directory as an earlier crawl), you can set the `crawl_time` spider argument.

kingfisher_scrapy/base_spider.py
```diff
@@ -75,6 +84,14 @@ def __init__(self, sample=None, note=None, from_date=None, until_date=None, late
     def from_crawler(cls, crawler, *args, **kwargs):
         spider = super(BaseSpider, cls).from_crawler(crawler, *args, **kwargs)
 
+        if spider.custom_collection_data_version:
+            try:
+                spider.custom_collection_data_version = datetime.strptime(spider.custom_collection_data_version,
+                                                                          '%Y-%m-%d')
```
jpmckinney, line 90: Since the crawl directory includes hours, minutes, seconds, we should include those in the input format.

That way, if you do one crawl normally, then decide you want to write more to that directory (/ collection), you can just specify the time in full without renaming the directory or updating the collection's data_version in Kingfisher Process.

kingfisher_scrapy/extensions.py
yolile: I changed the `get_start_time` method itself to use `crawl_time` or `self.crawler.stats.get_value('start_time')` accordingly, so we don't need to change it everywhere

PR #466, opened 2020-07-24
Generalize latestreleasedate pipeline to pluck any data from each publisher

jpmckinney:
- Refactor Pluck (LatestReleaseDate) item pipeline to be more generic
- Log the spiders that will be run by the pluck (latestreleasedate) command
- Add `package_pointer`, `release_pointer`, `truncate` and remove `latest` spider arguments
- Add `skip_pluck` (skip_latest_release_date) to georgia_records, mexico_nuevo_leon_records
- Rename classes, variables, settings and filenames accordingly
- Fix line lengths

It works with, for example:

```
scrapy pluck --package-pointer /publicationPolicy
scrapy pluck --release-pointer /date --truncate 10
```

I haven't added any new tests, however.


yolile:
I love the new command name! 
I'm not sure if we should create a new file depending on the arguments, eg, pluck-release-date.csv pluck-package-publicationPolicy.csv etc, and I'm also not sure if the "skip" option will be always valid for both (and all) the cases the same.

yolile:
I tried the new command and it works perfectly, but maybe we should add `skip_pluck` to `scotland_proactis` as it is not working now

jpmckinney:
> I'm not sure if we should create a new file depending on the arguments, eg, pluck-release-date.csv pluck-package-publicationPolicy.csv etc

Good idea, I'll add that.

> and I'm also not sure if the "skip" option will be always valid for both (and all) the cases the same.

I'm not sure, but for now the same reasons ("already covered" or file type not supported) work for our two use cases (release date and publication policy).

> I tried the new command and it works perfectly, but maybe we should add `skip_pluck` to `scotland_proactis` as it is not working now

Was Proactis fixed in #468?

yolile:
> Was Proactis fixed in #468?

Nop, scotland_public_contracts was fixed there, scotland_proactis is not working at all now.

jpmckinney:
Ah, okay. I think if a spider is broken, we don't need to add a `skip_pluck`. For now, we only need a `skip_pluck` if the source is already covered (to make the command complete earlier) or if the pluck framework doesn't yet support that type of spider (e.g. archive files, JSON Lines).

jpmckinney:
@yolile Thanks for approval! Can you merge and deploy?

yolile:
deployed!

Issue #407, opened 2020-06-01
Add a generic PeriodicSpider class


jpmckinney:
* [ ] moldova_old
* [ ] nepal_portal
* [ ] uruguay_historical

jpmckinney:
We also have some monthly spiders like scotland. We can perhaps have a combined solution (`PeriodicSpider`).

yolile:
closed by #467

PR #467, opened 2020-07-27
Add PeriodicalSpider

romifz:
Ref #407


kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
+    The ``pattern`` should include a placeholder for a year or month-year parameter. With the year parameter, an int is
+    passed. If the year-month parameter is used, a ``Date`` instance is passed. Example:
+
+    .. code-block: python
+
+        url = 'http://comprasestatales.gub.uy/ocds/rss/{0.year:d}/{0.month:02d}'
+
+    When the ``sample`` option is used, the latest year or month of data is retrieved.
+    """
+    VALID_DATE_FORMATS = {'year': '%Y', 'year-month': '%Y-%m'}
```
yolile, line 459: should we remove 'year-month' from `base_spider` `VALID_DATE_FORMATS` ?

kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
             raise MissingNextLinkError('next link not found on the first page: {}'.format(response.url))
+
+
+class PeriodicalSpider(SimpleSpider):
+    """
+    This class helps to crawl urls that receive a year (YYYY) or a month and year (YYYY-mm) as parameters. To use it:
+
+    1. Extend from ``PeriodicalSpider``.
+    1. Set the ``date_format`` attribute if it's not defined already. Valid values are 'year' and 'year-month'.
+    1. Set a ``start`` year or month-year.
```
yolile, line 445:
```suggestion
    1. Set a ``start`` year or year-month.
```

kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
+    1. Optionally, set a ``stop`` year or month-year. If absent, ``stop`` defaults to the current year or month-year.
```
yolile, line 446:
```suggestion
    1. Optionally, set a ``stop`` year or year-month. If absent, ``stop`` defaults to the current year or year-month.
```

kingfisher_scrapy/base_spider.py
yolile: @romifz

kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
+    .. code-block: python
+
+        url = 'http://comprasestatales.gub.uy/ocds/rss/{0.year:d}/{0.month:02d}'
+
+    When the ``sample`` option is used, the latest year or month of data is retrieved.
+    """
+    VALID_DATE_FORMATS = {'year': '%Y', 'year-month': '%Y-%m'}
+
+    def __init__(self, *args, **kwargs):
+        self.date_format_key = self.date_format
```
yolile, line 462: @romifz is this needed? cant we use `self.date_format` directly?

kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
+        else:
+            self.stop = datetime.today()
+
+    @classmethod
+    def get_default_until_date(cls, spider):
+        return spider.stop.strftime(spider.date_format)
+
+    def start_requests(self):
+
+        start = self.start if not self.exists('from_date') else self.from_date
```
yolile, line 486: @romifz maybe we can use `default_from_date` and `default_until_date` instead of creating `start` and `end` and dont reimplement the logic here?

kingfisher_scrapy/base_spider.py
```diff
@@ -429,3 +434,77 @@ def next_link(self, response):
...
+        self.date_format_key = self.date_format
+        super().__init__(*args, **kwargs)
+
+        if hasattr(self, 'start_requests_callback'):
+            self.start_requests_callback = getattr(self, self.start_requests_callback)
+        else:
+            self.start_requests_callback = self.parse
+
+        if not isinstance(self.start, DateClass):
+            self.start = datetime.strptime(str(self.start), self.date_format)
```
yolile, line 471: this can be always a str instead of a integer, and use `from_date` and `until_date` directly instead

kingfisher_scrapy/base_spider.py
romifz: I'm not sure. I think it should be possible to use the same format for filtering in other spiders, although I think we don't have a case, don't we?

kingfisher_scrapy/base_spider.py
romifz: In this case yes, because internally `self.date_format` is replaced by the date format it represents in `VALID_DATE_FORMATS` e.g. "year" gets replaced by "%Y". I prefer to have the original name in a separate variable.

kingfisher_scrapy/base_spider.py
romifz: I think we can

kingfisher_scrapy/base_spider.py
romifz: Sure

kingfisher_scrapy/base_spider.py
yolile: No, I think that all of the spiders which uses 'year-month' extends from `PeriodicalSpider` now

romifz: Ok, no problem, I'll remove the option.

kingfisher_scrapy/base_spider.py
```diff
@@ -115,17 +116,25 @@ def from_crawler(cls, crawler, *args, **kwargs):
             if not spider.from_date:
                 # Default to `default_from_date` class attribute.
                 spider.from_date = spider.default_from_date
+                if isinstance(spider.from_date, str):
+                    # convert to date format, if needed
+                    spider.from_date = datetime.strptime(spider.from_date, spider.date_format)
+            else:
```
yolile, line 122: should we remove the `else` statement and always convert the date to datetime? to avoid duplicating `spider.from_date = datetime.strptime(spider.from_date, spider.date_format)`

kingfisher_scrapy/base_spider.py
```diff
@@ -435,3 +448,82 @@ def next_link(self, response, **kwargs):
...
+        else:
+            self.start_requests_callback = self.parse
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super(SimpleSpider, cls).from_crawler(crawler, *args, **kwargs)
+
+        if not spider.from_date:
+            spider.from_date = spider.default_from_date
+            if isinstance(spider.from_date, str):
```
yolile, line 491: we are already doing this in the base class, arent we?

kingfisher_scrapy/spiders/scotland_base.py
```diff
@@ -1,51 +1,42 @@
...
-from kingfisher_scrapy.util import date_range_by_month, parameters
+from kingfisher_scrapy.base_spider import PeriodicalSpider
+from kingfisher_scrapy.util import parameters
 
 
-class ScotlandBase(SimpleSpider):
+class ScotlandBase(PeriodicalSpider):
     default_from_date = '2019-01'
     date_format = 'year-month'
+    default_until_date = date.today()
```
yolile, line 10: the base class already does this, doesn't it?

kingfisher_scrapy/base_spider.py
romifz: Yes, but the conversion in the base class happens only if the user specifies `from_date` or `until_date`.

kingfisher_scrapy/spiders/scotland_base.py
romifz: Yes, we can remove this.

kingfisher_scrapy/base_spider.py
romifz: This if-else block is used to know whether `from_date` was passed as a parameter or not. If we eliminate the `else` a bad configuration in `default_from_date` by a subclass may raise a `SpiderArgumentError`.

kingfisher_scrapy/base_spider.py
yolile: I think you can do something like  
https://github.com/open-contracting/kingfisher-collect/blob/726b8d0b4cd8d49b44d55b325e846fde34f4dbc9/kingfisher_scrapy/spiders/paraguay_dncp_base.py#L39-L43
to avoid the validation/conversion duplication

romifz: This would have the same issue as the previous comment, a bad configuration in the spider subclass may rise a `SpiderArgumentError`. Since Scrapy does not provide a way to know if a property is a parameter set by the user (at least I couldn't find a way), I think it is best to never set `from_date` and `until_date` in a superclass like `PeriodicalSpider`.

I'm thinking that the best solution would be to have a class property to indicate the base class that we need a time interval to be set, like `require_time_period` or something similar. Does this make sense?

yolile: I think that as `default_*_date` variables are hardcoded variables we shouldn't have bad configurations here...

romifz: That's fair.

kingfisher_scrapy/base_spider.py
yolile: should we remove the else then? (following the same logic that there are not going to be bad configurations in hardcoded default dates)

yolile: @romifz not sure if you have seen this!

romifz: Sorry, I'll check if we can remove the if-else block.

yolile (approved):

yolile:
closes #407

PR #480, opened 2020-09-01
Add get_parameter_value method

### LINKED ISSUE #478: Refactoring: Extract get_parameter from armenia and portugal
armenia.py has `get_offset`. portugal_base.py has very similar code. We can extract a common method that accepts a query string parameter name as an argument (in case we want to use it for something other than `'offset'` in the future).

aguilerapy:
closes #478

### DIFF
--- a/kingfisher_scrapy/util.py
+++ b/kingfisher_scrapy/util.py
@@ -165,7 +165,10 @@ def grouper(iterable, n, fillvalue=None):
     return itertools.zip_longest(*args, fillvalue=fillvalue)
 
 
-def get_parameter_value(url, parameter_name):
+def get_parameter_value(url, key):
+    """
+    Returns the value of the given ``key`` in ``url``.
+    """
     query = parse_qs(urlsplit(url).query)
-    if parameter_name in query:
-        return int(query[parameter_name][0])
+    if key in query:
+        return int(query[key][0])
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/util.py
```diff
@@ -163,3 +163,9 @@ def default(obj):
 def grouper(iterable, n, fillvalue=None):
     args = [iter(iterable)] * n
     return itertools.zip_longest(*args, fillvalue=fillvalue)
+
+
+def get_parameter_value(url, parameter_name):
+    query = parse_qs(urlsplit(url).query)
+    if parameter_name in query:
```
yolile, line 170:
```suggestion
    if key in query:
```

kingfisher_scrapy/util.py
```diff
@@ -163,3 +163,9 @@ def default(obj):
 def grouper(iterable, n, fillvalue=None):
     args = [iter(iterable)] * n
     return itertools.zip_longest(*args, fillvalue=fillvalue)
+
+
+def get_parameter_value(url, parameter_name):
```
yolile, line 168:
```suggestion
def get_parameter_value(url, key):
    """
    Returns the value of the given ``key`` in ``url``.
    """
```

kingfisher_scrapy/util.py
```diff
@@ -163,3 +163,9 @@ def default(obj):
...
+    query = parse_qs(urlsplit(url).query)
+    if parameter_name in query:
+        return int(query[parameter_name][0])
```
yolile, line 171:
```suggestion
        return int(query[key][0])
```

yolile (approved):

PR #481, opened 2020-09-01
Add new Dominican Republic scraper

### LINKED ISSUE #472: Dominican Republic API
http://148.101.176.123:48080/ocdsdr/docs#/

**Link to CRM issue**
https://crm.open-contracting.org/issues/5913#change-49248

**Priority**
Anytime

aguilerapy:
closes #472.

### DIFF
--- a/kingfisher_scrapy/spiders/dominican_republic_portal.py
+++ b/kingfisher_scrapy/spiders/dominican_republic_portal.py
@@ -1,0 +1,40 @@
+import scrapy
+
+from kingfisher_scrapy.base_spider import LinksSpider
+from kingfisher_scrapy.util import parameters
+
+
+class DominicanRepublicPortal(LinksSpider):
+    """
+    API documentation
+      http://148.101.176.123:48080/ocdsdr/docs
+    Spider arguments
+      sample
+        Downloads the first release package returned by the main endpoint.
+      from_date
+        Download only data from this date onward (YYYY-MM-DD format).
+        If ``until_date`` is provided, defaults to '2018-01-01'.
+      until_date
+        Download only data until this date (YYYY-MM-DD format).
+        If ``from_date`` is provided, defaults to today.
+    """
+    name = 'dominican_republic_portal'
+    data_type = 'release_package'
+    default_from_date = '2018-01-01'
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/spiders/dominican_republic_portal.py
```diff
@@ -0,0 +1,40 @@
...
+      sample
+        Downloads the first release package returned by the main endpoint.
+      from_date
+        Download only data from this date onward (YYYY-MM-DD format).
+        If ``until_date`` is provided, defaults to '2018-01-01'.
+      until_date
+        Download only data until this date (YYYY-MM-DD format).
+        If ``from_date`` is provided, defaults to today.
+    """
+    name = 'dominican_republic_portal'
```
yolile, line 21:
```suggestion
    name = 'dominican_republic_api'
```

kingfisher_scrapy/spiders/dominican_republic_portal.py
```diff
@@ -0,0 +1,40 @@
...
+    def start_requests(self):
+        url = 'http://148.101.176.123:48080/ocdsdr/api/v1/releases'
+        if self.from_date and self.until_date:
+            url = url + '/byDatesBetween/{}/{}'.format(
+                self.from_date.strftime('%Y-%m-%d'),
+                self.until_date.strftime('%Y-%m-%d')
+            )
+        yield scrapy.Request(url, meta={'file_name': 'page-1.json'}, callback=self.parse_response)
+
+    def parse_response(self, response):
```
yolile, line 35: the spider is not working as you are not returning anything if the response is 200 here. Also, this should be already covered by the use of `handle_http_error` so this callback is not necessary

aguilerapy: Yes, I accidentally deleted a line in that function to continue the scraper.

I wanted to give more information to the user in case they enter a date range where there is no data, otherwise, we only show this message:

imagen

yolile: I think that 404 error is clear enough but we can add that message to `handle_http_error` if `from_date` and/or `until_date` are used, in another issue

PR #495, opened 2020-09-17
Update ONCAE scraper

### LINKED ISSUE #429: Honduras ONCAE: Update sample to download a package for each system
ONCAE data comes from three different modules/systems:

* HonduCompras 1.0 - Módulo de Difusión de Compras y Contrataciones
* Módulo de Difusión Directa de Contratos
* Catálogo Electrónico

The sample parameter should download a package for each one, since the fields available may change depending of the source.

aguilerapy:
closes #429

### DIFF
--- a/kingfisher_scrapy/spiders/honduras_oncae.py
+++ b/kingfisher_scrapy/spiders/honduras_oncae.py
@@ -30,9 +30,9 @@ class HondurasONCAE(CompressedFileSpider):
     download_timeout = 900
 
     @classmethod
-    def from_crawler(cls, crawler, *args, **kwargs):
-        spider = super().from_crawler(crawler, *args, **kwargs)
-        if hasattr(spider, 'system') and spider.system not in spider.systems:
+    def from_crawler(cls, crawler, system=None, *args, **kwargs):
+        spider = super().from_crawler(crawler, system=system, *args, **kwargs)
+        if system and spider.system not in spider.systems:
             raise scrapy.exceptions.CloseSpider('Specified system is not recognized')
         return spider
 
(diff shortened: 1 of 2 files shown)


aguilerapy:
A new argument `system` was also added, which allows us to choose from which system the data will be taken.

kingfisher_scrapy/spiders/honduras_oncae.py
```diff
@@ -28,9 +45,18 @@ def start_requests(self):
...
-            urls = [urls[0]]
         for url in urls:
+            path, file = split(urlparse(url).path)
+            current_system = path.replace('/datosabiertos/', "")
+            if self.system and current_system != self.system:
+                continue
+            if self.sample:
+                if systems_flags[current_system]:
+                    continue
+                if next((system for system in systems_flags if not system), False):
```
yolile, line 58: @aguilerapy I find this a little bit hard to read, as `systems_flags` is a dict of booleans and as you are using `system` as variable name and there is a `self.system` variable too. Maybe it would be simpler to remove items from `self.systems` when you process a file for that system. eg: 
``` javascript
            if self.sample:
                # if we already downloaded a package for all the available systems
                if not self.systems:
                    return
                # if we already processed a file for the current system
                if current_system not in self.systems:
                    continue
                self.systems.remove(current_system)
```
or a set of downloaded_systems: 

``` javascript
        downloaded_systems = set()
        for url in urls:
            path, file = split(urlparse(url).path)
            current_system = path.replace('/datosabiertos/', "")
            if self.system and current_system != self.system:
                continue
            if self.sample:
                # if we already downloaded a package for all the available systems
                if downloaded_systems == self.systems:
                    return
                # if we already processed a file for the current system
                if current_system in downloaded_systems:
                    continue
                # add the current system to the set of downloaded_systems
                downloaded_systems.add(current_system)
```
and maybe you can rename `self.systems` to `available_systems`

yolile (changes requested):

kingfisher_scrapy/spiders/honduras_oncae.py
yolile: @aguilerapy I think that the second approach will be better as it doesn't remove items from the class variable, sorry!!

yolile (approved):

Issue #470, opened 2020-07-28
Allow users to specify any query string parameters


jpmckinney:
#346 is about adding a few common spider arguments, so that a user can set `-a from_date=2001-01-01` and not have to learn each API's parameter for that concept.

However, some users might know an API's parameters and want to use them.

With this feature, the user can add multiple spider arguments like `-a qs=offset:123456789 -a qs=buyer:xyz` and then `offset=123456789&buyer=xyz` would be added to the URLs in `start_requests`.

yolile:
closed by #504

PR #504, opened 2020-09-22
Add query parameters to `start_requests`' URL

romifz:
ref #470 

When a custom parameter is specified twice, Scrapy returns the last specified value only. So, in the example given in #470 

```-a qs=offset:123456789 -a qs=buyer:xyz```

Scrapy would return `qs=buyer:xyz`. The implementation proposed here can take more than one parameter using `,` as separator:

```-a qs=offset:123456789,buyer:xyz```

### CODE REVIEW

**Review by yolile — CHANGES_REQUESTED** [review]
@romifz I tested it with Paraguay and it works perfectly! 
The only problem is when the value itself has a comma, eg ` scrapy crawl paraguay_dncp_releases -a qs=buyer.name:"Ministerio de Urbanismo, Vivienda y Habitat"` the resulting URL has the parameter with the value just before the comma eg:
`https://contrataciones.gov.py/datos/api/v3/doc/search/processes?tipo_fecha=fecha_release&fecha_desde=2010-01-01T00%3A00%3A00&fecha_hasta=2020-09-22T09%3A51%3A01&buyer.name=Ministerio+de+Urbanismo&page=4`

and, could you add the new argument's documentation at `BaseSpider` please?

**Comment by romifz** [comment]
@yolile turns out that scrapy removes the quotes around the parameter values :(

I've modified the method to accept commas escaped using `\`, so for your example you can use: 

```
scrapy crawl paraguay_dncp_releases -a qs=buyer.name:"Ministerio de Urbanismo\, Vivienda y Habitat"
```

yolile (approved):

Issue #284, opened 2020-02-04
uk_fts: Update for new API responses


jpmckinney:
Their API now responds with a release package with a `links` field, like `georgia_records`, `georgia_releases`, `moldova_records`, `moldova_releases` and maybe others.

Instead of copy-pasting this code to `uk_fts`, we should instead create a base class (e.g. LinksSpider), so that we can centralize the logic for handling all sources that follow this API pattern. `LinksSpider` can go in `base_spider.py`.

PR #522, opened 2020-10-14
Add Sample pipeline

### LINKED ISSUE #84: Add "has sample mode" option per source

yolile:
closes #84


yolile:
@jpmckinney the pipeline plus the use of `s CLOSESPIDER_ITEMCOUNT=1 -s CONCURRENT_ITEMS=1` did the trick and this is working. However, I was wondering how could be the best way to document the use of these parameters. Should I update all the spiders that have the "sample" argument and say that these fields must be used with the `-s` option?

jpmckinney:
Instead of asking people to set those two, maybe we can write a `sample` command that sits in front of the `crawl` command and adds those arguments? For Scrapyd, I guess people will still have to set the arguments.

jpmckinney:
I think we can also remove all the code that uses `self.sample`

jpmckinney:
We might also refactor some of the pluck code. From #284 

> We can maybe extract the logic for counting items from `LatestReleaseDate` into this new pipeline. The `self.processed` logic in `LatestReleaseDate` essentially tests for `CLOSESPIDER_ITEMCOUNT=1` via presence of the spider name in a set. We can have more generic logic that tests for whatever number is set by `CLOSESPIDER_ITEMCOUNT`.

The `pluck` command can set `CLOSESPIDER_ITEMCOUNT` and `CONCURRENT_ITEMS`.

yolile:
> I think we can also remove all the code that uses self.sample.

Yes! I will do that too, I just created the draft PR to discuss/try the best solution before doing further changes

I changed the pipeline now to don't have to use any settings at all, and it seems to work without having to use extra arguments or commands, just with our normal "sample" argument

jpmckinney:
Nice!

yolile:
Do you think we should add a "number_of_samples" argument to indicate how many items the user wants to download or we should fix a number?

jpmckinney:
If we're keeping the `sample` boolean, then I think it's best to change it to an integer. (For backwards compatibility, we can just transform 'true' to 1.)

kingfisher_scrapy/extensions.py
```diff
@@ -126,6 +126,12 @@ def from_crawler(cls, crawler):
 
         return extension
 
+    def _is_sample(self, spider):
+        sample = False
+        if spider.sample:
+            sample = True
+        return sample
```
jpmckinney, line 133: Replace calls to `_is_sample` with `bool(spider.sample)`

kingfisher_scrapy/pipelines.py
```diff
@@ -54,7 +54,10 @@ def __init__(self):
     def process_item(self, item, spider):
         if not spider.sample:
             return item
-        if self.item_count >= 1:
+        # Drop FileError items, so that we keep trying to get data.
+        if not isinstance(item, (File, FileItem)):
+            raise DropItem()
+        if self.item_count >= spider.sample:
             spider.crawler.engine.close_spider(spider, 'closespider_itemcount')
```
jpmckinney, line 61: Should we use a different reason for close_spider, instead of reusing an existing extension's reason? (It could be `'sample'`)

kingfisher_scrapy/spiders/afghanistan_releases.py
```diff
@@ -12,7 +12,7 @@ class AfghanistanReleases(SimpleSpider):
       https://ocds.ageops.net/
     Spider arguments
       sample
-        Downloads the first release returned by the release endpoint of the API.
+        Sets the number of releases to download.
```
jpmckinney, line 15: Since this is now available for all spiders, we can instead edit `updatedocs.py` to mention it at the top of the file, and then remove it from all individual spiders' docs.

jpmckinney (commented):
Looks great! Just a few comments.

kingfisher_scrapy/spiders/afghanistan_releases.py
yolile: I have done that for most of the spiders now, but left some comments on some of them, eg the ones that download the entire file but sends only X to kingfisher.

jpmckinney (approved):

PR #525, opened 2020-10-20
Add --sample option to command

jpmckinney:
This builds in #522 

I wanted to download a single file from all spiders, and this was the easiest way to do it!

Since this changes the semantics of the dryrun command, I'll rename it to `crawlall`.


jpmckinney:
@yolile I can rebase this branch after #522 is reviewed and merged. (I think you're still adding commits?)

yolile:
@jpmckinney sorry, no, I haven't realized that I didn't request your review again, I did it now

PR #543, opened 2020-11-03
Add 'scraping one day' condition in LinksSpider

### LINKED ISSUE #412: Change LinksSpider to error if the next link isn't found on the first page

aguilerapy:
closes #412

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -73,6 +73,12 @@ class BaseSpider(scrapy.Spider):
         if self.query_string_parameters and hasattr(self, 'start_requests'):
             self.start_requests = add_query_string(self.start_requests, self.query_string_parameters)
 
+        self.filter_arguments = {
+            'from_date': from_date,
+            'until_date': until_date,
+        }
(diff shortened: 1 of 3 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -444,6 +444,9 @@ def next_link(self, response, **kwargs):
         if url:
             return self.build_request(url, formatter=self.next_page_formatter, **kwargs)
 
+        if self.from_date and self.until_date and (self.until_date - self.from_date).days <= 1:
```
yolile, line 447:
```suggestion
        if self.from_date and self.until_date:
```
the just one day was an example, I think we should check only if these parameters are set. @jpmckinney what do you think?

jpmckinney: The `MissingNextLinkError` exception is just to help developers identify when their spider is broken. I think we can return if any filters are set (not just dates).

yolile (approved):

jpmckinney (approved):

Issue #486, opened 2020-09-09
Add number of each type of item to crawl stats


jpmckinney:
It would be useful to track the number of file, fileitem, and fileerror items that have been sent to Kingfisher Process.

Kingfisher Archive can also use these stats to assess the error rate. (Open an issue there once this is closed.)

Related #411

yolile:
closded by #551

PR #551, opened 2020-11-06
Add stats about items sent to kingfisher process

yolile:


jpmckinney:
#486

jpmckinney:
Oops, I described something different than what I wanted in #486. Instead, what is desired is just a count of items that make it through the pipeline. So, we can put this in a new extension.

yolile:
@jpmckinney sure, no worries, so the new extension should count each type of item and then in `spider_close` log the counts?

jpmckinney:
I think it can just `inc_value` on the `item_scraped` handler.

yolile:
ready to review again @jpmckinney

kingfisher_scrapy/extensions.py
```diff
@@ -100,6 +101,29 @@ def _write_file(self, path, data):
                 json.dump(data, f)
 
 
+class KingfisherItemCount:
+    def __init__(self, stats):
+        self.stats = stats
+        for item_type in KingfisherItem.__subclasses__():
+            self.stats.set_value(self._get_count_type_name(item_type), 0)
```
jpmckinney, line 108: I don't think we need to set initial values for `inc_value` to work (and if values aren't present in the log, we know to interpret it as `0`).

kingfisher_scrapy/extensions.py
```diff
@@ -100,6 +101,29 @@ def _write_file(self, path, data):
...
+    @classmethod
+    def from_crawler(cls, crawler):
+        extension = cls(crawler.stats)
+        crawler.signals.connect(extension.item_scraped, signal=signals.item_scraped)
+        return extension
+
+    def item_scraped(self, item, spider):
+        self.stats.inc_value(self._get_count_type_name(item))
+
+    def _get_count_type_name(self, item):
```
jpmckinney, line 119: If we remove the above code, we can inline this method into `item_scraped`. We can also simplify the method, since the item will always be an object.

kingfisher_scrapy/extensions.py
```diff
@@ -100,6 +101,29 @@ def _write_file(self, path, data):
...
+        class_ = type(item)
+        # the item can be a class or an object, if it is already a class we dont need its type()
+        if isinstance(item, ItemMeta):
+            class_ = item
+        return f'kingfisher_{class_.__name__.lower()}_count'
```
jpmckinney, line 124: We can omit the `kingfisher_` prefix, since Scrapy doesn't define any item types, besides base classes.

kingfisher_scrapy/extensions.py
yolile: Yes, we don't need to but I just thought that maybe having the values explicit to 0 instead of checking if the values are present or not would be better, but sure, I will remove the initialization then

jpmckinney (approved):
I made one further simplification. The extension is nice and small!

PR #555, opened 2020-11-11
Update Honduras scrapers

aguilerapy:
#346

### DIFF
--- a/kingfisher_scrapy/spiders/honduras_portal_bulk_files.py
+++ b/kingfisher_scrapy/spiders/honduras_portal_bulk_files.py
@@ -68,10 +68,10 @@ class HondurasPortalBulkFiles(PeriodicSpider):
                 for system in self.oncae_systems:
                     if self.system and system != self.system:
                         continue
-                    yield self.pattern.format(f"{self.available_publishers.get(publisher)}_"
-                                              f"{self.oncae_systems.get(system)}_{date.year}_{date.month:02d}.json")
+                    yield self.pattern.format(f"{self.available_publishers[publisher]}_"
+                                              f"{self.oncae_systems[system]}_{date.year}_{date.month:02d}.json")
             else:
-                yield self.pattern.format(f"{self.available_publishers.get(publisher)}_"
+                yield self.pattern.format(f"{self.available_publishers[publisher]}_"
                                           f"{date.year}_{date.month:02d}.json")
 
     def get_formatter(self):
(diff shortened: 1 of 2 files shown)


aguilerapy:
I ran `scrapy updatedocs` but it doesn't modify anything.

jpmckinney:
That's expected. The `autoclass` directive will render the new content.

kingfisher_scrapy/spiders/honduras_portal_bulk_files.py
```diff
@@ -1,54 +1,76 @@
...
-    def from_crawler(cls, crawler, publisher=None, *args, **kwargs):
-        spider = super().from_crawler(crawler, publisher=publisher, *args, **kwargs)
-        if publisher and spider.publisher not in spider.publishers:
+    def from_crawler(cls, crawler, publisher=None, system=None, *args, **kwargs):
+        spider = super().from_crawler(crawler, publisher=publisher, system=system, *args, **kwargs)
+        if publisher and spider.publisher not in spider.available_publishers:
             raise SpiderArgumentError(f'spider argument `publisher`: {spider.publisher!r} not recognized')
 
-        spider.publisher_name = spider.publishers.get(publisher)
+        if system:
```
yolile, line 53: we cant use `system` with `publisher = sefin` right? maybe we should add a validation for that too

aguilerapy: Do you mean raise an error for that case? Because I added `spider.publisher = 'oncae'` (in [line 56](https://github.com/open-contracting/kingfisher-collect/pull/555/files#diff-892fde9e96bc97af7d74aefb493db6a8dc4e5823cfe1e5e974b30a08eb0d68e0R56)) if `system` is given.

yolile: > Do you mean raise an error for that case? 

Yes, I think we should raise a SpiderArgumentError

yolile (dismissed):

kingfisher_scrapy/spiders/honduras_portal_bulk_files.py
```diff
@@ -1,54 +1,78 @@
...
-            yield self.build_request(url, formatter=components(-1))
+
+            if publisher == 'oncae':
+                for system in self.oncae_systems:
+                    if self.system and system != self.system:
+                        continue
+                    yield self.pattern.format(f"{self.available_publishers.get(publisher)}_"
+                                              f"{self.oncae_systems.get(system)}_{date.year}_{date.month:02d}.json")
+            else:
+                yield self.pattern.format(f"{self.available_publishers.get(publisher)}_"
```
jpmckinney, line 74:
```suggestion
                    yield self.pattern.format(f"{self.available_publishers[publisher]}_"
                                              f"{self.oncae_systems[system]}_{date.year}_{date.month:02d}.json")
            else:
                yield self.pattern.format(f"{self.available_publishers[publisher]}_"
```

yolile (approved):

Issue #292, opened 2020-02-06
Use Twisted for Kingfisher Process API requests


jpmckinney:
Presently, we add a new dependency (`requests`) to make requests in `kingfisher_process.py`. But, Scrapy is already built on Twisted. We should use that instead.

*(Deleted long, old proposal, to avoid confusion)*

jpmckinney:
Once this and #293 are done, we can remove `requests` as a dependency (also requires updates to documentation).

jpmckinney:
With Scrapy 2, it might be possible to write shorter / cleaner code: https://docs.scrapy.org/en/latest/topics/coroutines.html

jpmckinney:
Not sure if perhaps we can use `spider.crawler.engine.download` like in this example: https://docs.scrapy.org/en/latest/topics/item-pipeline.html#take-screenshot-of-item

I think extensions can also return Deferred's, in case that can be useful.

jpmckinney:
Aha! Some Scrapy examples use [treq](https://github.com/twisted/treq), which has a requests-like API, but uses Twisted.

jpmckinney:
We can use `await` with treq, so that the KFP request doesn't block: https://docs.scrapy.org/en/latest/topics/coroutines.html

jpmckinney:
* treq doesn't support proxies (which is fine: we disabled them for requests): https://treq.readthedocs.io/en/release-20.9.0/#feature-parity-with-requests
* Its `_convert_files` method requires a different order to `files` tuples: https://treq.readthedocs.io/en/release-20.9.0/_modules/treq/client.html?highlight=_convert_files

yolile:
- Its `_converted` method supports only: tuples, lists, bytes or text parameters (not booleans eg sample=true): https://treq.readthedocs.io/en/release-20.9.0/_modules/treq/multipart.html?highlight=_converted

yolile:
closed by #559

PR #559, opened 2020-11-16
[WIP] use treq instead of requests

yolile:
ref #292

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,8 @@ jobs:
     - run: curl -s -S --retry 3 $BASEDIR/tests/install.sh | bash -
     - run: pip install -r requirements_dev.txt
     - run: curl -s -S --retry 3 $BASEDIR/tests/script.sh | bash -
-    - run: pytest --cov kingfisher_scrapy
+    # Specify tests/ directory to avoid "ModuleNotFoundError: No module named 'sphinx_rtd_theme'" from docs/conf.py.
+    - run: pytest --cov kingfisher_scrapy tests
       if: always()
     - env:
         COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     - run: curl -s -S --retry 3 $BASEDIR/tests/install.sh | bash -
     - run: pip install -r requirements_dev.txt
     - run: curl -s -S --retry 3 $BASEDIR/tests/script.sh | bash -
-    - run: python -m pytest --cov kingfisher_scrapy tests
+    - run: pytest --cov kingfisher_scrapy
       if: always()
     - env:
         COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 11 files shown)


yolile:
@jpmckinney I was trying to use `treq.post` instead of `request.post` but, there is a problem. It seems like sometimes, if the file sent to kingfisher is big then the spider finished and is closed before the request is complete and I get:

> twisted.internet.task.TaskStopped
Unhandled error in Deferred:
2020-11-16 15:01:23 [twisted] CRITICAL: Unhandled error in Deferred:
Traceback (most recent call last):
Failure: twisted.web._newclient.ResponseNeverReceived: [<twisted.python.failure.Failure twisted.internet.error.ConnectionLost: Connection to the other side was lost in a non-clean fashion: Connection lost.>]

And the other problem is that the `end_collection` request could be sent before all the `create_file` ones

jpmckinney:
@yolile How did you test this? Just so we're using the same crawl command when debugging.

jpmckinney:
Also, were you sending the requests to a local copy of Kingfisher Process that you were running?

jpmckinney:
@yolile You have to return the deferreds :) (my latest commit). Please document in #292 why we need to call `str()` on booleans. The last thing to do is update tests to check for the log messages. You might need to use these helpers: https://treq.readthedocs.io/en/release-20.9.0/testing.html

yolile:
@jpmckinney great! I was testing it with a local deployment of kingfisher and with `uruguay_releases -a sample=1` and with `canada_buyandsell`, both work now! I will update the remaining tests, and I already commented in #292 the reason why we use str and not booleans

jpmckinney:
@yolile: I got the tests passing. I was trying to use async/await, but Twisted requires Awaitables to be converted to Deferreds, so I just used inlineCallbacks.

jpmckinney:
We can improve the tests by using more of https://treq.readthedocs.io/en/release-20.9.0/testing.html (instead of stubbing treq.post), but if you can manually test the integration with KFP, I'm fine with leaving as-is.

jpmckinney:
We could have gotten the tests to pass without `inlineCallbacks`. What was missing was a callback for treq.post, which would have returned the response: https://treq.readthedocs.io/en/release-20.9.0/#post

jpmckinney:
Hmm, I'm realizing that we mock the return value of `treq.post` as a response object, but its real return value is a Deferred. 

Also, I'm not sure that https://treq.readthedocs.io/en/release-20.9.0/testing.html helps us. First, `StubTreq` doesn't support the files argument: https://treq.readthedocs.io/en/release-20.9.0/api.html#stubtreq-objects so we can't use it to test `create_file`. Secondly, it seems like the requests need to be directly consumed by the test, which would require adding hooks deep into our code.

jpmckinney:
My latest commit reverts my changes in 56b9825812192a9a434686ec86af0da350a84374 and uses the pytest-twisted module to inline the callback *in the test* (my earlier commit inlined it in the extension, which was incorrect).

The tests require httpbin.org, which can fail unexpectedly. Unfortunately, vcrpy doesn't support Twisted (yet): https://github.com/kevin1024/vcrpy/issues/538

There might be a way to enable Scrapy's HttpCacheMiddleware in tests and store the results like vcrpy would.

yolile:
@jpmckinney this looks great, and I tested it locally and works like a charm!

jpmckinney (approved):

Issue #329, opened 2020-03-04
Send only basic types to Kingfisher Process


jpmckinney:
Until https://github.com/open-contracting/kingfisher-process/issues/209 is closed, if a data source embeds OCDS data in a new place, we need to add a new data type to Kingfisher Process, and then use that `data_type` in the spider. After that issue is closed, we can just specify e.g. a `root_path` to indicate where the OCDS data is embedded.

~~To avoid the overhead of creating more data types, some spiders index to the OCDS data, and then use `save_data_to_disk` to save just the embedded OCDS data. This isn't ideal, as we're not storing the additional content, which might be relevant.~~

The spiders that do this are:

* [x] honduras_portal_records (avoids a `record_list_in_results` type)
* [x] honduras_portal_releases (avoids `release_list_in_results`)
* [x] mexico_quien_es_quien (avoids `record_package_list_in_data`)
* [x] openopps (avoids `release_package_in_json_in_list_in_results`)

-----

Instead of closing the issue in Process, in https://github.com/open-contracting/kingfisher-collect/issues/329#issuecomment-730846155 and https://github.com/open-contracting/kingfisher-collect/pull/565#discussion_r527157819, I suggest addressing these issues in Collect. The above spiders should configure a `data_pointer` (or `root_path`) and an item pipeline should index to that part of the data.

jpmckinney:
Spiders that use `parse_json_lines` and some spiders that inherit from `ZipSpider` also don't send the full response to Kingfisher Process.

jpmckinney:
Same for `parse_json_array`.

jpmckinney:
Although there has been no discussion in #482, I doubt the data outside the OCDS data is very useful. In any case, Kingfisher Process doesn't know what to do with it, and just discards it (even now). So, I'm changing the definition of this issue to instead do the 'root path' processing within Kingfisher Collect, and to only send OCDS data to Kingfisher Process.

yolile:
Updated list of spiders that don't use record, release, record_package or release_package `data_type`:

- [ ] ArgentinaVialidad `release_package_list`
- [ ] BoliviaAgetic `release_list`
- [ ] ColombiaBulk `release_in_Release`
- [ ] KenyaMakueni `release_package_list`
- [ ] Kosovo `release_list`
- [ ] MexicoAdministracionPublicaFederal `record_package_list_in_results`
- [ ] MexicoQuienEsQuien `record_package_list`
[...]

PR #565, opened 2020-11-18
Add Filter pipeline

### LINKED ISSUE #346: Implement from_date and until_date spider arguments

aguilerapy:
#692

### DIFF
--- a/kingfisher_scrapy/pipelines.py
+++ b/kingfisher_scrapy/pipelines.py
@@ -6,6 +6,7 @@ import pkgutil
 import tempfile
 
 import jsonpointer
+from dateutil.parser import parse
 from flattentool import unflatten
 from jsonschema import FormatChecker
 from jsonschema.validators import Draft4Validator, RefResolver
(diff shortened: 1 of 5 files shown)


jpmckinney:
I changed the PR description, since it doesn't close the issue (there still need to be updates to DNCP docstrings).

kingfisher_scrapy/pipelines.py
```diff
@@ -160,6 +161,94 @@ def process_item(self, item, spider):
...
+
+    def _date_filter(self, release):
+        release_date = parse(release['date'], ignoretz=True)
+        return self.filter_arguments['from_date'] <= release_date <= self.filter_arguments['until_date']
+
+    def process_item(self, item, spider):
+        if spider.from_date and spider.until_date:
+            self.filter_arguments['from_date'] = spider.from_date
+            self.filter_arguments['until_date'] = spider.until_date
+            self.filter_functions.append(self._date_filter)
```
yolile, line 213: this can be done in the `open_spider` method

kingfisher_scrapy/pipelines.py
```diff
@@ -160,6 +161,94 @@ def process_item(self, item, spider):
...
+
+        # More filter functions can be added here
+
+        if self.filter_functions:
```
yolile, line 217: we also need to check if the scraper has not already the filter at the API level and if the Item is a `File` or `FileItem`

kingfisher_scrapy/pipelines.py
```diff
@@ -160,6 +161,94 @@ def process_item(self, item, spider):
...
+        for record in records:
+            if 'releases' in record:
+                record['releases'] = self._filter_releases(record['releases'])
+                if record['releases']:
+                    filtered_records.append(record)
+
+        return filtered_records
+
+    def _filter_releases(self, releases):
+        filtered_releases = []
```
yolile, line 197: I think we can simplify this a little bit by using something like:
```python
    def _filter_releases(self, releases):
        filtered_releases = releases
        for function in self.filter_functions:
            filtered_releases = list(filter(function, filtered_releases))

        return filtered_releases
```

kingfisher_scrapy/pipelines.py
```diff
@@ -160,6 +161,94 @@ def process_item(self, item, spider):
...
+        for release in releases:
+            for function in self.filter_functions:
+                if function(release):
+                    filtered_releases.append(release)
+
+        return filtered_releases
+
+    def _date_filter(self, release):
+        release_date = parse(release['date'], ignoretz=True)
```
yolile, line 206: do we need to use an extra library `(dateutil)` to parse the dates?

kingfisher_scrapy/pipelines.py
```diff
@@ -160,6 +161,94 @@ def process_item(self, item, spider):
...
+            self.filter_arguments['from_date'] = spider.from_date
+            self.filter_arguments['until_date'] = spider.until_date
+            self.filter_functions.append(self._date_filter)
+
+        # More filter functions can be added here
+
+        if self.filter_functions:
+            data = json.loads(item['data'])
+
+            if 'package_list' in item['data_type']:
```
yolile, line 220: I'm wondering if we could have a dict with the key and functions per data_type in order to avoid some many ifs blocks, eg:
```python
        data_types_functions = {'package_list_in_results': {'key': 'results', 'function': self._filter_packages},
                                'release_package_list': {'key': None, 'function': self._filter_packages},
                                  ....
```

so then we could do something like:

```python
        if self.filter_functions:
            data = json.loads(item['data'])
            object_type = data_types_functions[item['data_type']]
            if object_type['key']:
                data[object_type['key']] = object_type['function'](data[object_type['key']], item['data_type'])
                self._check_data_filtered(data[object_type['key']])
            else:
                data = object_type['function'](data, item['data_type'])
                self._check_data_filtered(data)

            item['data'] = json.dumps(data)

        return item
```

what do you think @jpmckinney ?

jpmckinney: This is related to https://github.com/open-contracting/kingfisher-collect/issues/329

The proposal was to have a `root_path` class attribute (same name as in OCDS Kit) to point the root package/release/record. In most cases this should ideally be `''` (i.e. top-level), but for `package_list_in_results` it would be `results.item` for ijson or `/results` for json_pointer (though I guess we'd need a second class attribute to indicate that it's an array). We've already sort of done this in SimpleSpider with `data_pointer` and `data_type`.

We can clean this up on Kingfisher Collect's side, such that `data_type` is always one of the canonical OCDS formats (`release_package`, etc.). Then, in the Kingfisher Process extension, we'd map to an appropriate `data_type`.

If we want to do that to clean up this PR, I suggest starting a new PR focused only on cleaning up `data_type`. Once that's done, we can continue this PR and clean up this method.

yolile: I think this is a good time to clean up `data_type` and it is ok to do it in Kingfisher Collect. So, let's work on #329 first and then come back to this PR

jpmckinney: This PR is now unblocked, now that #329 is fixed.

yolile:
I will close this PR as there are too many changes now and it is easier to start over

PR #566, opened 2020-11-24
Update docstrings

aguilerapy:
#346

### DIFF
--- a/kingfisher_scrapy/spiders/paraguay_dncp_records.py
+++ b/kingfisher_scrapy/spiders/paraguay_dncp_records.py
@@ -8,7 +8,10 @@ class ParaguayDNCPRecords(ParaguayDNCPBaseSpider):
     Spider arguments
       from_date
         Download only records from this date onward (YYYY-MM-DDTHH:mm:ss format).
-        If ``from_date`` is not provided, defaults to '2010-01-01T00:00:00'.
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/spiders/paraguay_dncp_records.py
```diff
@@ -8,7 +8,10 @@ class ParaguayDNCPRecords(ParaguayDNCPBaseSpider):
     Spider arguments
       from_date
         Download only records from this date onward (YYYY-MM-DDTHH:mm:ss format).
-        If ``from_date`` is not provided, defaults to '2010-01-01T00:00:00'.
```
yolile, line 11: I think this is still true, as `date_required` is true in `ParaguayDNCPBase` https://github.com/open-contracting/kingfisher-collect/blob/2b6bc9439d86306e9261e3239d34cf9f32e3844c/kingfisher_scrapy/spiders/paraguay_dncp_base.py#L26

same for `until_date` and `paraguay_dncp_releases`

yolile (changes requested):
and could you remove 
```python
    Spider arguments
      sample
        Downloads the first file in the downloads page.
```
from `BoliviaAgetic` ? otherwise `scrapy checkall` complains about it

yolile (approved):

PR #572, opened 2020-12-02
Add KingfisherTransformMiddleware and update affected spiders

### LINKED ISSUE #329: Send only basic types to Kingfisher Process

### LINKED ISSUE #449: pluck: Support for ZIP or RAR files
Let's see if we can support at least some of the publishers of ZIP/RAR files.

### LINKED ISSUE #450: pluck: Digiwhist support
Are the JSON Lines files completely unordered? Is there not something we can do?

### LINKED ISSUE #471: CompressedFileSpider: process compressed files is useless when KingfisherProcess is not configured
Currently, this class saves the original file to disk and yields the compressed files to the KingfisherProcess extension, so if the user is not using Kingfisher Process this process is useless. 

So maybe we can ask if `KINGFISHER_API_URI` is set and if it is not, skip this process.

### LINKED ISSUE #573: Consider merging logic for data_pointer and root_path
Follow-up to #572

### LINKED ISSUE #574: Consider merging logic for compressed_file_format and file_format
Follow-up to #572

yolile:
closes #329, closes #449, closes #450, closes #471, closes #573, closes #574

(this will also solve getting only samples for compressed files and also being able to pluck them)

@jpmckinney this is a draft, and before updating the documentation and tests I would like to read your opinion.


kingfisher_scrapy/base_spider.py
```diff
@@ -34,13 +33,15 @@ class BaseSpider(scrapy.Spider):
...
-    MAX_RELEASES_PER_PACKAGE = 100
     VALID_DATE_FORMATS = {'date': '%Y-%m-%d', 'datetime': '%Y-%m-%dT%H:%M:%S'}
 
     ocds_version = '1.1'
     date_format = 'date'
     date_required = False
     unflatten = False
+    root_path = None
+    # override this if the file is in json_line format
+    file_format = None
```
jpmckinney, line 44: Do we need both `file_format` and `compressed_file_format` or can we collapse them?

jpmckinney (commented):
Looks good! Some comments.

The `data_pointer` logic is doing something similar to `root_path`. Should we switch it to `root_path`?

kingfisher_scrapy/base_spider.py
```diff
@@ -390,20 +348,19 @@ def parse(self, response):
                 basename += '.json'
 
             data = archive_file.open(filename)
+            package = None
+            # for compressed_file_format == 'release_package' we need to read the file twice: once to extract the
+            # package metadata and then to extract the releases themselves
+            if self.compressed_file_format == 'release_package':
+                package = archive_file.open(filename)
```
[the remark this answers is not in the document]
yolile: This was already like that before, we only used it for `argentina_buenos_aires` where they JSON file is too big to be processed by kingfisher process

kingfisher_scrapy/items.py
```diff
@@ -22,11 +22,8 @@ class File(KingfisherItem):
     files_store = scrapy.Field()
 
 
-class FileItem(KingfisherItem):
+class FileItem(File):
```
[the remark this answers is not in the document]
yolile: you are totally right, I hadn't thought about that, I will repeat the fields instead

kingfisher_scrapy/base_spider.py
yolile: I added that just for `digiwhist_base` that currently doesn't extend from `CompressedFileSpider`

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +79,82 @@ def process_request(request, spider):
...
+            yield spider.build_file_item(number=number, file_name=file_name, url=url, data=data,
+                                         data_type=data_type, encoding=encoding)
+
+    def _parse_json_lines(self, spider, data, *, file_name='data.json', url=None, data_type=None, encoding='utf-8'):
+        for number, line in enumerate(data, 1):
+            if isinstance(line, bytes):
+                line = line.decode(encoding=encoding)
+            yield from self._parse_json_array(spider, line, line, file_name=file_name, url=url, data_type=data_type,
+                                              encoding=encoding)
+            return
```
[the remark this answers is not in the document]
yolile: ups my bad, I left that while doing some tests. When we are parsing compressed files, there can be a lot of lines and most of them are dropped if we use the sample option. Should we add a condition here asking if the sample option is set to avoid dropping a lot of items?

kingfisher_scrapy/spiders/afghanistan_records.py
```diff
@@ -16,6 +16,7 @@ class AfghanistanRecords(SimpleSpider):
     name = 'afghanistan_records'
     data_type = 'record'
     skip_pluck = 'Already covered (see code for details)'  # afghanistan_releases
+    root_path = ''
```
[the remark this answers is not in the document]
yolile: The middleware asks for` root_path is not None` to transform the item. In this case, as the data is a record, we still want to transform it into a record_package, so I set the root_path to '' for the middleware process it and transform it into a record package. Maybe we can instead add a more explicit field to indicate that the middleware should process it (or maybe just ask if `data_type` is not `release_package` or `record_package`)

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +79,82 @@ def process_request(request, spider):
...
+    def _parse_json_array(self, spider, package_data, list_data, *, file_name='data.json', url=None, data_type=None,
+                          encoding='utf-8'):
+
+        list_type = 'releases'
+        if 'record' in data_type:
+            list_type = 'records'
+
+        package = self._get_package_metadata(package_data, list_type, data_type, spider.root_path)
+        # we change the data_type into a valid one:release_package or record_package
+        data_type = data_type if 'package' in data_type else f'{data_type}_package'
```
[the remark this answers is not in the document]
yolile: > Also, can this code be reached by items with data_type zip, rar or tar.gz?

Actually, I think that we don't have that kind of items anymore, I will update the item schema to reflect that

kingfisher_scrapy/middlewares.py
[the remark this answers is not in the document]
jpmckinney: Yes, that sounds like a good optimization.

kingfisher_scrapy/spiders/afghanistan_records.py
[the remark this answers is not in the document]
jpmckinney: Yes, I think it will be clearer if the default `root_path` is `''`, and if we use the condition you suggest, to decide whether to process in the middleware.

kingfisher_scrapy/base_spider.py
jpmckinney: Okay, we can maybe look at combining them in a follow-up PR.

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +79,97 @@ def process_request(request, spider):
...
+            package[list_type] = filter(None, items)
+            data = json.dumps(package, default=util.default)
+            yield spider.build_file_item(number=number, file_name=file_name, url=url, data=data,
+                                         data_type=data_type, encoding=encoding)
+
+    def _parse_json_lines(self, spider, data, *, file_name='data.json', url=None, data_type=None, encoding='utf-8'):
+        for number, line in enumerate(data, 1):
+            # to avoid dropping items
+            if spider.sample and number > spider.sample:
+                return
```
jpmckinney, line 156: If there are many items in `result` in `process_spider_output`, then I guess we'll still do extra processing, but it will at least be limited.

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +79,97 @@ def process_request(request, spider):
...
+            list_type = 'releases'
+
+        package = self._get_package_metadata(package_data, list_type, data_type, spider.root_path)
+        # we change the data_type into a valid one:release_package or record_package
+        if data_type in ('release', 'record'):
+            data_type = f'{data_type}_package'
+            key = spider.root_path
+        # if the array is a list of packages then we point to the releases or records items
+        else:
+            key = f'{spider.root_path}.{list_type}.item'
```
jpmckinney, line 139: What if the `spider.root_path` is `''`? Maybe something like `'.'.join(list(filter(None, [spider.root_path, list_type, 'item'])))`

jpmckinney (commented):
I reviewed my earlier comments. I'll read the diff in full later.

kingfisher_scrapy/middlewares.py
yolile: I think that the combination of `spider.root_path = ''` and `data_type in ('release_package', 'record_package')` can not reach to this block

kingfisher_scrapy/base_spider.py
jpmckinney: Follow-up issue is here: https://github.com/open-contracting/kingfisher-collect/issues/574

kingfisher_scrapy/middlewares.py
jpmckinney: Okay, let's either add a comment, or use the safe code.

kingfisher_scrapy/middlewares.py
```diff
@@ -90,58 +90,74 @@ class KingfisherTransformMiddleware:
...
 
         # we yield a release o record package with a maximum of self.MAX_RELEASES_PER_PACKAGE releases or records
-        for number, items in enumerate(util.grouper(ijson.items(list_data, spider.root_path),
+        for number, items in enumerate(util.grouper(ijson.items(list_data, key),
                                                     self.MAX_RELEASES_PER_PACKAGE), 1):
+            # to avoid dropping items
```
jpmckinney, line 144: Is this to avoid dropping items, or rather to avoid reading a large file, when we know the later items will be dropped?

jpmckinney:
I've read the diff now. Looks good! Just those few comments left.

kingfisher_scrapy/middlewares.py
yolile: both! I will update the comment

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +79,102 @@ def process_request(request, spider):
...
+            # if it is a compressed file and the file does'nt need any transformation
+            if compressed_file and spider.compressed_file_format is None:
+                item['data'] = data.read()
+                yield item
+            # if it is a compressed file or regular file but as json_lines
+            elif spider.file_format or (compressed_file and spider.compressed_file_format == 'json_lines'):
+                yield from self._parse_json_lines(spider, data, **kwargs)
+            # otherwise is must be a release or record package or a list of them
+            else:
+                yield from self._parse_json_array(spider, package, data, **kwargs)
```
jpmckinney, line 122: I think this comment isn't quite right. One or more of the following is true: ``data_type`` is release or record, or ``compressed_file_format`` is release_package, or ``root_path`` is non-empty.

As such, I think the method name should also be changed (or maybe we should have different methods).

For example, is afghanistan_records/releases processed correctly? https://ocds.ageops.net/api/record/5ed2a62c4192f32c8c74a4e5 returns a single record. We can wrap it in a simple package, and then return it as a file, instead of as a file item. This can be done in a separate, simpler method.

That method is very hard to analyze, because it also gets called from _parse_json_lines. I think we should have multiple methods for specific stages. For example, maybe:

* items = _split_file_data(...): if it's json lines, yields each line, otherwise yields data back
* items = _apply_root_path(items, ...): if root_path is set, returns ijson.items for each item, otherwise yields each item back
* items = _add_package(items, ...): if data_type is release or record, adds a simple package to each item, otherwise yields each item back
* _resize_packages: does the self.MAX_RELEASES_PER_PACKAGE logic if needed (I think only absolutely necessary when compressed_file_format = 'release_package')

jpmckinney (commented):
One comment to refactor to make code easier to analyze.

kingfisher_scrapy/middlewares.py
jpmckinney: Actually, maybe we should have separate, simpler middlewares (similar idea to the above)?

kingfisher_scrapy/base_spider.py
yolile: @jpmckinney I ended up removing it now

yolile:
@jpmckinney now we have a different middleware per step, what do you think?

tests/middlewares/test_kingfisher_transform_middleware.py
```diff
@@ -0,0 +1,156 @@
...
+    (FileError({
+        'file_name': 'test',
+        'url': 'http://test.com',
+        'errors': ''
+    }))
+]
+
+
+@pytest.mark.parametrize('item', items)
+def test_yield_items(item):
```
jpmckinney, line 33: Should this method be further parametrized? As I understand, `item` isn't modified in place, so each section of code within the test is independent.

jpmckinney (commented):
I merged master. The test files pass when run in isolation, but not when run all together. I assume something is maintaining state between test files causing the failure. Can you have a look?

jpmckinney:
Nevermind - tests were failing locally for a different reason that in GitHub Actions.

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +78,120 @@ def process_request(request, spider):
...
+        if 'record' in data_type:
+            list_type = 'records'
+        else:
+            list_type = 'releases'
+        return list_type
+
+
+class KingfisherTransformResizePackageMiddleware:
+    """
+     If spider.root_path is set, calls ijson.items and yields each value, otherwise yields each item back
```
jpmckinney, line 170: This isn't the correct docstring.

tests/middlewares/test_kingfisher_transform_middleware.py
jpmckinney: Done

jpmckinney:
An important fix is https://github.com/open-contracting/kingfisher-collect/pull/572/commits/b2beb4daaf865715776c2f8dda76f1f4c8c5d6f8. Before that, `uk_contracts_finder`, for example, would yield duplicate items, because the root path middleware only changed the data, but not the file_name or number, which are part of the unique key for the File or FileItem.

kingfisher_scrapy/middlewares.py
jpmckinney: Fixed

jpmckinney (dismissed):

jpmckinney (dismissed):

yolile:
@jpmckinney thank you for your changes and improvements!

jpmckinney:
> Note that I had to change test_data_types - please review
@yolile

jpmckinney:
@yolile More specifically, I was wondering about: https://github.com/open-contracting/kingfisher-collect/pull/572/commits/b2beb4daaf865715776c2f8dda76f1f4c8c5d6f8#diff-1071f4e84a30adb8a2b145f878256523e64f9aa931b5a41dceec4320f9d3420fL75-R78

Originally, the `response_mock`'s data key was changed, and then submitted along with the original `item` to `add_package_middleware`. In the new code, the `response_mock` is not changed, and the item returned by the `root_path_middleware` middleware is submitted to `add_package_middleware`. I just wanted to check that this was correct. I wasn't sure why the `response_mock` was changed in the old code.

yolile:
> Originally, the response_mock's data key was changed, and then submitted along with the original item to add_package_middleware. In the new code, the response_mock is not changed, and the item returned by the root_path_middleware middleware is submitted to add_package_middleware. I just wanted to check that this was correct. I wasn't sure why the response_mock was changed in the old code.

@jpmckinney I think that was redundant and it is ok to remove it, as actually we don't use the response in the middleware, just `result`

jpmckinney:
@yolile Thanks! I've removed it, so I'll merge this once CI is done. 🎉

yolile:
@jpmckinney awesome, I'm so happy right now, look at all these closed issues, it's beautiful!

jpmckinney:
Same! ☺️

PR #597, opened 2021-01-15
portugal: Implement exponential backoff

### LINKED ISSUE #545: portugal_records and portugal_releases return only a subset of the data
According to the publisher, the API contains the same data as the bulk file (>1.5 million records) but `portugal_records` returns c. 500 records and `portugal_releases` returns only c. 4,000 releases.

I checked `collection_file` and there are a handful of errors reported, but otherwise the spider seems to have run successfully:

id | collection_id | filename | warnings | url | errors
-- | -- | -- | -- | -- | --
157,599,196 | 1,539 | offset-55.json | [NULL] | http://www.base.gov.pt/api/Record/GetRecords?offset=55 | {"http_code": 503}
157,598,197 | 1,539 | ocid-ocds-px0z7d-4457164-6424548-1.json | [NULL] | http://www.base.gov.pt/api/Record/GetRecordByOCID?ocid=ocds-px0z7d-4457164-6424548-1 | {"http_code": 503}
157,598,183 | 1,539 | ocid-ocds-px0z7d-4352773-6422690-1.json | [NULL] | http://www.base.gov.pt/api/Record/GetRecordByOCID?ocid=ocds-px0z7d-4352773-6422690-1 | {"http_code": 503}
157,598,175 | 1,539 | ocid-ocds-px0z7d-4376315-6417617-1.json | [NULL] | http://www.base.gov.pt/api/Record/GetRecordByOCID?ocid=ocds-px0z7d-4376315-6417617-1 | {"http_code": 503}
157,598,149 | 1,539 | ocid-ocds-px0z7d-4460028-6430538-1.json | [NULL] | http://www.base.gov.pt/api/Record/GetRecordByOCID?ocid=ocds-px0z7d-4460028-6430538-1 | {"http_code": 503}
157,597,864 | 1,542 | offset-404.json | [NULL] | http://www.base.gov.pt/api/Release/GetReleases?offset=404 | {"http_code": 429}

I'm following up separately with the publisher about the packaging and pagination issues.

yolile:
closes #545


yolile:
I tried this on the server https://collect.kingfisher.open-contracting.org/logs/kingfisher/portugal_releases/0ec81dd8576711eb85470c9d92c523cb.log and it was working but I think that now the site is completely down 
image

yolile:
I made some changes again and started another run, that is still running here: https://collect.kingfisher.open-contracting.org/logs/kingfisher/portugal_releases/c4d6f8ec599711eb85470c9d92c523cb.log

It is on page 702.601 so it looks promising

cc @duncandewhurst @jpmckinney

yolile:
It worked! https://collect.kingfisher.open-contracting.org/logs/kingfisher/portugal_releases/ec20e9da5a4611eb85470c9d92c523cb.log

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
...
+
     def start_requests(self):
         url = self.url
         if self.from_date and self.until_date:
             url = f'{url}&contractStartDate={self.from_date}&contractEndDate={self.until_date}'
         yield scrapy.Request(url, meta={'file_name': 'offset-1.json'})
+
+    def parse(self, response):
+        # after a undefined number of requests (around 36100?), the API returns 5XX errors.
+        # After waiting a few seconds the URL works again
```
jpmckinney, line 34: 180 seconds is a few minutes, not a few seconds.

I assume shorter wait times didn't work consistently?

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
...
+        if not self.is_http_success(response):
+            if self.number_of_retries < self.max_number_of_retries:
+                self.logger.info(f'Response status {response.status} waiting {self.wait_time} '
+                                 f'seconds before continue, attempt {self.number_of_retries}')
+                yield self.build_file_error_from_response(response)
```
jpmckinney, line 39: I think we only want to yield errors if retrying failed. Yielding errors tells Process/Archive how many requests never received a successful response.

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
...
+                time.sleep(self.wait_time)
```
jpmckinney, line 40: Sleep will block, but in Twisted/Scrapy we want to avoid blocking. Can you have a look at https://twistedmatrix.com/documents/13.1.0/core/howto/time.html ?

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
...
+                self.number_of_retries = self.number_of_retries + 1
```
jpmckinney, line 41:
```suggestion
                self.number_of_retries += 1
```

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
...
+                # if it fails again for the same request, we wait more
+                self.wait_time = self.wait_time * self.number_of_retries
+                yield scrapy.Request(response.request.url, dont_filter=True, meta=response.request.meta)
+            else:
+                self.logger.info(f'Response status {response.status}, maximum attempts reached, giving up')
+                yield self.build_file_error_from_response(response)
```
jpmckinney, line 47: I assume we want to reset number_of_retries and wait_time here, too?

tests/test_spiders.py
```diff
@@ -33,6 +33,10 @@ def test_start_requests_http_error(spider_name):
             response = Response('http://example.com', status=555, request=request)
             items = list(callback(response))
 
+            # if all errors are allowed the spider is handling its own errors and could retry after an error
+            if spider.custom_settings and spider.custom_settings.get('HTTPERROR_ALLOW_ALL', False):
```
jpmckinney, line 37: Hmm, maybe test for the presence of a `number_of_retries` class attribute instead.

kingfisher_scrapy/spiders/portugal_base.py
jpmckinney: We could start with a shorter wait time, and do exponential backoff, e.g. `self.wait_time *= 2`, instead of linear backoff. Right now we try at 180, 360 ... 2700 (45 mins, total waits=6 hours), but we could do 1, 2, 4 ... 16384 (4.5 hours, total waits=9 hours). If we do exponential backoff, we should probably reduce to max retries (e.g. 8192=2.3 hours, total waits=4.5 hours).

yolile: > 180 seconds is a few minutes, not a few seconds.

You are right!

> I assume shorter wait times didn't work consistently?

Yes, I tried with 120 seconds first. I think that we can use the exponential approach as well, and limit the retries to 10?

jpmckinney: Sure - you can decide the initial wait time and the max retries, based on your experience with the API.

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -11,11 +13,39 @@ class PortugalBase(LinksSpider):
     download_delay = 1
     # The API returns 503 error sometimes
     custom_settings = {
-        'RETRY_TIMES': 10
+        'RETRY_TIMES': 10,
+        'AUTOTHROTTLE_ENABLED': True,
+        'AUTOTHROTTLE_TARGET_CONCURRENCY': 0.5,
```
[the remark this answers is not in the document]
yolile: nop, removed

tests/test_spiders.py
yolile: done

kingfisher_scrapy/spiders/portugal_base.py
yolile: I changed it now (although it wasn't so simple as the spider closes when idle so I added a signal to avoid that) see https://stackoverflow.com/questions/46698333/deferred-requests-in-scrapy

kingfisher_scrapy/spiders/portugal_base.py
yolile: ok sure, changed!

jpmckinney:
1. I think the number of retries and `wait_time` needs to go in the request meta. `CONCURRENT_REQUESTS_PER_DOMAIN` is 2 (and we might change it in future), so `number_of_retries` (renamed to `retries`) on the class will be shared by multiple requests, whereas we want one for each request. See Scrapy's RetryMiddleware's code, for example.
2. Are you sure we want to always raise DontCloseSpider in `spider_idle`? Won't that cause the spider to never close? The repo linked from the StackOverflow discussion suggests that it isn't necessary since Scrapy 1.0: https://github.com/ArturGaspar/scrapy-delayed-requests
1. I assume we still want `download_delay = 1, but just checking.

yolile:
> I think the number of retries and wait_time needs to go in the request meta. CONCURRENT_REQUESTS_PER_DOMAIN is 2 (and we might change it in future), so number_of_retries (renamed to retries) on the class will be shared by multiple requests, whereas we want one for each request. See Scrapy's RetryMiddleware's code, for example.

Or we can set CONCURRENT_REQUESTS_PER_DOMAIN to 1 for this spider, but agree that a generic solution is better, I'll change that.

> Are you sure we want to always raise DontCloseSpider in spider_idle? Won't that cause the spider to never close? The repo linked from the StackOverflow discussion suggests that it isn't necessary since Scrapy 1.0: https://github.com/ArturGaspar/scrapy-delayed-requests

Yes, the spider is closed if not. But you are right, we need to add a stop condition otherwise the spider will never be closed. I will think what the condition could be.

> I assume we still want download_delay = 1, but just checking.

With the retry approach, it works even without it

yolile:
> Yes, the spider is closed if not. But you are right, we need to add a stop condition otherwise the spider will never be closed. I will think what the condition could be.

Update: it works if used as https://github.com/ArturGaspar/scrapy-delayed-requests: as a middleware

yolile:
@jpmckinney could you check if you are happy with the middleware so I will add the tests for it?

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -7,15 +7,48 @@
...
         url = self.url
         if self.from_date and self.until_date:
             url = f'{url}&contractStartDate={self.from_date}&contractEndDate={self.until_date}'
+
         yield scrapy.Request(url, meta={'file_name': 'offset-1.json'})
+
+    def parse(self, response):
+
+        retries = response.request.meta.get('retries', 0)
+        wait_time = response.request.meta.get('delay_request', self.initial_wait_time)
```
jpmckinney, line 33: I think the code is easier to follow if `delay_request` is renamed to `wait_time`

kingfisher_scrapy/middlewares.py
```diff
@@ -74,3 +76,18 @@ def process_request(request, spider):
...
+    """
+    Downloader middleware that allows for delaying a request by a set 'delay_request' number of seconds.
+
+    A delayed request is useful when an API fails and works again after waiting a few minutes.
+    """
+    def process_request(self, request, spider):
+        delay = request.meta.get('delay_request', None)
+        if delay:
+            d = Deferred()
+            reactor.callLater(delay, d.callback, None)
```
jpmckinney, line 92: The request is ignored except to get the delay?

In case it's relevant, Scrapy has the methods `defer_succeed`, `defer_result` and `mustbe_deferred`: https://github.com/scrapy/scrapy/blob/master/scrapy/utils/defer.py And this is some slightly different code: https://github.com/scrapy/scrapy/issues/802#issuecomment-498749742

yolile: > The request is ignored except to get the delay?

Yes, that is how the downloader middleware works https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#scrapy.downloadermiddlewares.DownloaderMiddleware

kingfisher_scrapy/middlewares.py
```diff
@@ -6,6 +6,8 @@
 from twisted.internet import reactor
 from twisted.internet.defer import Deferred
 
+WAIT_META = 'wait_time'
```
jpmckinney, line 9: I prefer to just use the same string inline instead of having a constant (Scrapy does the same for its common strings). Our tests should have sufficient coverage to catch any incorrect strings.

kingfisher_scrapy/middlewares.py
jpmckinney: I'm having trouble understanding the downloader middleware. The docs say:

> Any of the downloader middleware methods may also return a deferred.

However, the docs for `process_request` don't say what happens if a deferred is returned (which is what we return if there is a delay; otherwise we return None). If I read the code, it looks like _InvalidOutput is raised if the return is not None, a Response, or a Request. Can you add a test for this middleware to guarantee it works?

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -1,21 +1,55 @@
...
-    # The API returns 503 error sometimes
+
+    # We will wait 1, 2, 4, 8, 16 minutes (31 minutes total).
+    max_retries = 6
+    initial_wait_time = 30
+
     custom_settings = {
-        'RETRY_TIMES': 10
+        'DOWNLOADER_MIDDLEWARES': {
+            'kingfisher_scrapy.middlewares.DelayedRequestsMiddleware': 543,
```
jpmckinney, line 20: Since the middleware does nothing unless wait_time is set, we can add this to settings.py

jpmckinney (commented):
> > I assume we still want download_delay = 1, but just checking.
> 
> With the retry approach, it works even without it

Ok, let's remove it.

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -1,21 +1,55 @@
...
 
 class PortugalBase(LinksSpider):
     default_from_date = '2010-01-01'
     next_page_formatter = staticmethod(parameters('offset'))
-    # The API return 429 error after a certain number of requests
+    # The API otherwise returns HTTP 429.
     download_delay = 1
-    # The API returns 503 error sometimes
+
+    # We will wait 1, 2, 4, 8, 16 minutes (31 minutes total).
```
jpmckinney, line 14: You changed the `max_retries` and `initial_wait_time`, so you need to update this comment.

yolile: it is still true as I multiply/increase them before sending them to the middleware

jpmckinney: Ah, right. Thanks!

jpmckinney: Actually, I'm not sure:

```
In [1]: retries = 0

In [2]: wait_time = 30

In [3]: while retries < 6:
   ...:     retries += 1
   ...:     wait_time *=2
   ...: 

In [4]: wait_time
Out[4]: 1920
```

1920 is 32 mins. I'll commit a small change.

yolile: @jpmckinney thanks!

kingfisher_scrapy/middlewares.py
yolile: I'm trying to find a way to test the middleware (more than just calling the process_request method) to test what happens with the deferred, although I already test it with portugal_releases and it seems to be working

jpmckinney: I haven't looked at them, but maybe there's inspiration from Scrapy's retry middleware tests? https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware_retry.py

yolile: they only test process_request and process_response responses directly :( the best thing I found was https://github.com/scrapy/scrapy/blob/master/tests/test_downloadermiddleware.py#L36 and used it

tests/middlewares/test_delayed_request_middleware.py
```diff
@@ -0,0 +1,33 @@
+from scrapy import Request
+from scrapy.core.downloader import DownloaderMiddlewareManager
+from twisted.internet.defer import Deferred
+from twisted.trial.unittest import TestCase
+
+from kingfisher_scrapy.middlewares import DelayedRequestMiddleware
+from tests import spider_with_crawler
+
+
+def mock_download_func(spider, request):
```
jpmckinney, line 10: Move this inside the test method.

tests/middlewares/test_delayed_request_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+    return request
+
+
+def test_delayed_middleware():
```
jpmckinney, line 14: Can you split the test into separate methods? It seems to have a few different scenarios. Ideally, we have one scenario per test, so that one scenario's failure doesn't shadow the others. If the scenarios are very similar, you can use `pytest.mark.parametrize`.

jpmckinney: We can update this part of the style guide: https://ocp-software-handbook.readthedocs.io/en/latest/python/code.html#tests

tests/middlewares/test_delayed_request_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+    spider = spider_with_crawler()
+    delay_middleware = DelayedRequestMiddleware()
+    request = Request(f'http://example.com', meta=None)
```
jpmckinney, line 17: Don't need f-string if no substitutions.

jpmckinney (approved):

jpmckinney:
I just made some code style changes to the tests. Approved!

Issue #590, opened 2021-01-14
Support both rar and zip files in a single spider


aguilerapy:
For example, the Dominican Republic portal has `rar` and `zip` files: https://www.dgcp.gob.do/datos-abiertos/estandar-mundial-ocds/.

- [x] Add functionality to support rar and zip files in a single scraper
- [x] Add that functionality to the Dominican Republic spider
- [ ] Update date filter to consider 2020 in Dominican Republic spider

yolile:
closed by #602

PR #602, opened 2021-01-26
Support rar and zip files in the same spider

yolile:
ref #590

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -189,7 +189,7 @@ class BaseSpider(scrapy.Spider):
         :rtype: scrapy.Request
         """
         file_name = formatter(url)
-        if not file_name.endswith(('.json', '.zip', '.xlsx', '.csv')):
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -367,14 +366,15 @@ def start_requests(self):
     encoding = 'utf-8'
     skip_pluck = 'Archive files are not supported'
     compressed_file_format = None
-    archive_format = 'zip'
     file_name_must_contain = ''
 
     @handle_http_error
     def parse(self, response):
+        archive_name, archive_format = os.path.splitext(response.request.meta['file_name'])
+        archive_format = archive_format.replace('.', '')
```
jpmckinney, line 374:
```suggestion
        archive_format = archive_format[1:].lower()
```

Works even if `archive_format = ''`. Handles a potential future case where the filename from the source is capitalized.

jpmckinney (approved):

PR #605, opened 2021-02-10
604 data type fixes

### LINKED ISSUE #604: Fixes after #572
- Costa Rica Poder Judicial Releases (reported by  @nativaldezt ) generates wrong JSON outputs
- Colombia Bulk doesn't work anymore

yolile:
closes #604

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -38,6 +38,8 @@ class BaseSpider(scrapy.Spider):
        If you need to set more arguments for the unflatten command, set a ``unflatten_args`` dict with them.
     -  If the data is not formatted as OCDS (record, release, record package or release package), set a ``root_path``
        class attribute to the path to the OCDS data.
+    -  If the JSON file is line-delimited and the root path is to a JSON array root_path_max_length, set a
+       ``root_path_max_length`` class attribute to the maximum length of the JSON array at the root path.
     -  If the data is line-delimited JSON, add a ``line_delimited = True`` class attribute.
 
     If ``date_required`` is ``True``, or if either the ``from_date`` or ``until_date`` spider arguments are set, then
@@ -343,7 +345,7 @@ class CompressedFileSpider(BaseSpider):
             if self.resize_package:
                 data = {'data': compressed_file, 'package': archive_file.open(filename)}
             else:
-                data = compressed_file
+                data = compressed_file.read()
 
             yield File({
                 'file_name': basename,
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -12,7 +12,7 @@ from rarfile import RarFile
 from kingfisher_scrapy import util
 from kingfisher_scrapy.exceptions import MissingNextLinkError, SpiderArgumentError, UnknownArchiveFormatError
 from kingfisher_scrapy.items import File, FileError
-from kingfisher_scrapy.util import add_query_string, handle_http_error, get_file_name_and_extension
+from kingfisher_scrapy.util import add_query_string, get_file_name_and_extension, handle_http_error
 
 browser_user_agent = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36'  # noqa: E501
 
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -12,7 +12,7 @@ from rarfile import RarFile
 from kingfisher_scrapy import util
 from kingfisher_scrapy.exceptions import MissingNextLinkError, SpiderArgumentError, UnknownArchiveFormatError
 from kingfisher_scrapy.items import File, FileError
-from kingfisher_scrapy.util import add_query_string, handle_http_error
+from kingfisher_scrapy.util import add_query_string, handle_http_error, get_file_name_and_extension
 
 browser_user_agent = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36'  # noqa: E501
 
@@ -316,8 +316,7 @@ class CompressedFileSpider(BaseSpider):
 
     @handle_http_error
     def parse(self, response):
-        archive_name, archive_format = os.path.splitext(response.request.meta['file_name'])
-        archive_format = archive_format[1:].lower()
+        archive_name, archive_format = get_file_name_and_extension(response.request.meta['file_name'])
 
         if archive_format == 'zip':
             cls = ZipFile
@@ -345,7 +344,7 @@ class CompressedFileSpider(BaseSpider):
             if self.resize_package:
                 data = {'data': compressed_file, 'package': archive_file.open(filename)}
             else:
-                data = compressed_file.read()
+                data = compressed_file
 
             yield File({
                 'file_name': basename,
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -80,7 +80,7 @@ class KingfisherFilesStore:
 
         file_name = item['file_name']
         if isinstance(item, FileItem):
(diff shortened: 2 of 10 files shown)


kingfisher_scrapy/extensions.py
```diff
@@ -80,7 +80,7 @@ def item_scraped(self, item, spider):
 
         file_name = item['file_name']
         if isinstance(item, FileItem):
-            file_name += f"-{item['number']}"
```
jpmckinney, line 83: Oops, I had changed it because I wanted to yield `basename-number.json`, but the old code would do `basename.json-number`. Can we `os.splitext` the extension, and then add it back? The number is secondary to the basename, so it'd be easier for a user to read filenames if they start with the basename.

jpmckinney: Can you also add a test for this behavior?

kingfisher_scrapy/base_spider.py
```diff
@@ -343,7 +345,7 @@ def parse(self, response):
             if self.resize_package:
                 data = {'data': compressed_file, 'package': archive_file.open(filename)}
             else:
-                data = compressed_file
+                data = compressed_file.read()
```
jpmckinney, line 348: Do we want to `read()` here, or should we just test for file-like objects in the middlewares (`if hasattr(obj, 'read')`). I'm thinking that reading a compressed JSON Lines file is going to consume a lot of memory.

tests/test_compressed_file_spider.py
```diff
@@ -28,7 +28,6 @@ def test_parse():
     assert item['url'] == 'http://example.com'
     assert item['data_type'] == 'release_package'
     assert item['encoding'] == 'utf-8'
-    assert 'package' not in item['data']
```
jpmckinney, line 31: If we remove this, can we have a more specific test for `item['data']`?

kingfisher_scrapy/extensions.py
yolile: Sure!

kingfisher_scrapy/base_spider.py
yolile: Yes, you are right, in the first version I had a CompressedFileTransform as well. Now I can add that condition in all the middlewares but maybe also add a new one eg CompressedFileTransform at the end again to just read the file if the item is still a `File`? otherwise for cases like Costa Rica, where the only transformation required is to decompress the files, the output will be wrong (as it is now, because now the `File` is not processed by any middleware and therefore it is not read either)

jpmckinney: Yes, sounds good!

tests/test_compressed_file_spider.py
yolile: ups, I forgot about this one, I can add these asserts again now

kingfisher_scrapy/extensions.py
```diff
@@ -80,14 +80,16 @@ def item_scraped(self, item, spider):
...
+            name, extension = get_file_name_and_extension(file_name)
+            file_name = f"{name}-{item['number']}.{extension}"
 
         path = os.path.join(name, spider.get_start_time('%Y%m%d_%H%M%S'), file_name)
 
         self._write_file(path, item['data'])
 
         item['path'] = path
         item['files_store'] = self.directory
+        item['file_name'] = file_name
```
jpmckinney, line 92: Hmm, this will change the file name that is sent to Kingfisher, which isn't what we want. I think to test the behavior, we should test whether the file exists.

Can you add a test case for this? I would have expected a test to fail because of this change.

jpmckinney (commented):
I changed the middleware to be more generic. Also, if in the future we add another middleware that yields Files from CompressedFileSpiders whose `data` is read, then this middleware won't try to read it a second time.

kingfisher_scrapy/extensions.py
yolile: Actually, I wasn't sure about this, if kingfisher process will fail or not if the same file name is passed, but now that I think about it, for theses cases kinfgisher process uses the `number` attribute, rigth? I will change this back then

jpmckinney: Yes, Kingfisher Process accepts either Files (unique file_name) or FileItems (unique file_name and number).

yolile: done!

kingfisher_scrapy/extensions.py
```diff
@@ -66,23 +66,24 @@ def from_crawler(cls, crawler):
...
+        If the item is a File or FileItem, writes its data to the filename in the crawl's directory.
 
         Returns a dict with the metadata.
         """
         if not isinstance(item, (File, FileItem)):
             return
 
         # The crawl's relative directory, in the format `<spider_name>[_sample]/<YYMMDD_HHMMSS>`.
-        name = spider.name
+        folder_name = spider.name
```
jpmckinney, line 77: In OCDS code, `directory` is preferred to `folder`, to match the Python naming convention.

tests/extensions/test_kingfisher_files_store.py
```diff
@@ -45,23 +45,27 @@ def test_item_scraped_with_build_file_from_response(sample, path, tmpdir):
...
-    ('true', os.path.join('test_sample', '20010203_040506', 'file.json')),
+    (None, os.path.join('test', '20010203_040506')),
+    ('true', os.path.join('test_sample', '20010203_040506')),
 ])
 @pytest.mark.parametrize('data', [b'{"key": "value"}', {"key": "value"}])
-def test_item_scraped_with_build_file(sample, path, data, tmpdir):
+@pytest.mark.parametrize('item,expected_file_name', [(File({'file_name': 'file.json', 'encoding': 'iso-8859-1'}),
+                                                      'file.json'),
+                                                     (FileItem({'number': 1, 'file_name': 'file.json'}),
+                                                      'file-1.json')])
```
jpmckinney, line 55: For readability, I prefer arrays like:

```
a very long line with not much room at the end for an array: [
    long item that can still fit on one line and is easy to read,
    another item,
]
```

rather than:

```
a very long line with not much room at the end for an array: [(long line that
                                                               wraps onto many
                                                               lines and is
                                                               hard to read)]
```

tests/extensions/test_kingfisher_files_store.py
```diff
@@ -45,23 +45,27 @@ def test_item_scraped_with_build_file_from_response(sample, path, tmpdir):
...
+def test_item_scraped_with_file_and_file_item(sample, path, data, tmpdir, item, expected_file_name):
```
jpmckinney, line 56: To avoid confusion with parametrize arguments, I prefer to put pytest fixtures at the end:

```suggestion
def test_item_scraped_with_file_and_file_item(sample, path, data, item, expected_file_name, tmpdir):
```

tests/extensions/test_kingfisher_files_store.py
```diff
@@ -45,23 +45,27 @@ def test_item_scraped_with_build_file_from_response(sample, path, tmpdir):
...
     spider = spider_with_files_store(tmpdir, sample=sample)
     extension = KingfisherFilesStore.from_crawler(spider.crawler)
-
-    item = spider.build_file(file_name='file.json', url='https://example.com/remote.json', data=data,
-                             data_type='release_package', encoding='iso-8859-1')
+    path = os.path.join(path, expected_file_name)
```
jpmckinney, line 59: Below you can do `with open(tmpdir.join(path, expected_file_name)) as f:` since the variable is only used once.

We might want to rename the `path` parametrize argument to `directory` for clarity, since it no longer includes the filename.

yolile: > Below you can do with open(tmpdir.join(path, expected_file_name)) as f: since the variable is only used once.

actually it is used twice, for the assert  too https://github.com/open-contracting/kingfisher-collect/blob/beddb8d8cfe5f6166263fecb5beec7a52451cf3d/tests/extensions/test_kingfisher_files_store.py#L66

jpmckinney: Aha! I missed that.

jpmckinney (approved):

Issue #371, opened 2020-04-27
Have consistent Python type for items' data


odscjames:
```
2020-04-27 17:50:34 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method FeedExporter.item_scraped of <scrapy.extensions.feedexport.FeedExporter object at 0x7fc2e8dbb860>>
Traceback (most recent call last):
  File "/home/ocdskfs/scrapyd/.ve/lib/python3.6/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/ocdskfs/scrapyd/.ve/lib/python3.6/site-packages/pydispatch/robustapply.py", line 55, in robustApply
    return receiver(*arguments, **named)
  File "/home/ocdskfs/scrapyd/.ve/lib/python3.6/site-packages/scrapy/extensions/feedexport.py", line 268, in item_scraped
    slot.exporter.export_item(item)
  File "/home/ocdskfs/scrapyd/.ve/lib/python3.6/site-packages/scrapy/exporters.py", line 93, in export_item
    data = self.encoder.encode(itemdict) + '\n'
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/home/ocdskfs/scrapyd/.ve/lib/python3.6/site-packages/scrapy/utils/serialize.py", line 36, in default
    return super(ScrapyJSONEncoder, self).default(o)
  File "/usr/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable
```

Seems to happen once per line, as it tries to send to server

Cyprus scraper

jpmckinney:
I think we just to do  in `parse_json_lines`:

```python
if isinstance(line, bytes):
  line = line.decode()
```

… since ZipFile.read returns bytes, not str. (We also call `parse_json_lines` after `tar.extractfile`, and I'm not sure if that returns bytes or str. Anyway, an if-statement seems safest.)

jpmckinney:
cc @aguilerapy

jpmckinney:
[...]

PR #609, opened 2021-02-16
kingfisher process api: convert file item data to str before sending …

### LINKED ISSUE #608: JSONDecodeError: Expecting value: line 1 column 1 (char 0)
Sentry Issue: [KINGFISHER-PROCESS-7Y](https://sentry.io/organizations/open-contracting-partnership/issues/2219589528/?referrer=github_integration)

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
(6 additional frame(s) were not displayed)
...
  File "flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "ocdskingfisherprocess/web/views_api_v1.py", line 160, in dispatch_request
    data = json.loads(request.form.get('data'))
  File "__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
```

yolile:
…it to process

closes #608

### DIFF
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -198,7 +198,7 @@ class KingfisherProcessAPI:
 
         if isinstance(item, FileItem):
             data['number'] = item['number']
-            if isinstance(item['data'], str):
+            if isinstance(item['data'], (str, bytes)):
                 data['data'] = item['data']
             else:
                 data['data'] = json.dumps(item['data'], default=util.default)
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -199,6 +199,10 @@ class KingfisherProcessAPI:
         if isinstance(item, FileItem):
             data['number'] = item['number']
             data['data'] = item['data']
+            if isinstance(item['data'], str):
+                data['data'] = item['data']
+            else:
+                data['data'] = json.dumps(item['data'], default=util.default)
 
             return self._request(spider, 'create_file_item', item['url'], data)
 
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -198,7 +198,6 @@ class KingfisherProcessAPI:
 
         if isinstance(item, FileItem):
             data['number'] = item['number']
-            data['data'] = item['data']
             if isinstance(item['data'], str):
                 data['data'] = item['data']
             else:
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/extensions.py
```diff
@@ -199,6 +199,10 @@ def item_scraped(self, item, spider):
         if isinstance(item, FileItem):
             data['number'] = item['number']
             data['data'] = item['data']
```
jpmckinney, line 201: Delete this line.

jpmckinney (dismissed):

jpmckinney:
Actually, can you look at https://github.com/open-contracting/kingfisher-collect/issues/371? The data can sometimes be bytes, instead of str.

yolile:
@jpmckinney hmm that issue is about Digiwhist but now Digiwhist pass through the line delimited middleware as well. But the last comment in the issue also mentions that bytes is preferred over str, so do you want me to change it to bytes instead?

jpmckinney:
I closed #371 as I don't think it's a general problem.

However, it led me to review how we use encode/decode/json.dump/json/load/bytes/str and I noticed a few things:

* I see one case in `mexico_quien_es_quien` where JSON data is dumped then encoded to bytes, which I don't think is necessary. And, actually, I think that spider could just use `root_path` instead of indexing to the `'data'` key.

Otherwise, most of the relevant code is in `middlewares.py`:

* ResizePackageMiddleware dumps the package, but I don't think this is necessary (it's done by the extensions). However, we'd need to deepcopy the package metadata if we remove that line.
* AddPackageMiddleware loads the data to add `{"releases":[...]}` around the data. Technically, we could just do this using an f-string if the input is a str or bytes.
* LineDelimitedMiddleware decodes the bytes to str, but I don't remember why this behavior was introduced in https://github.com/open-contracting/kingfisher-collect/commit/7d406748a36cf073f2cff93a80c23c324324d6f2

kingfisher_scrapy/extensions.py
```diff
@@ -198,7 +198,10 @@ def item_scraped(self, item, spider):
 
         if isinstance(item, FileItem):
             data['number'] = item['number']
-            data['data'] = item['data']
+            if isinstance(item['data'], str):
```
jpmckinney, line 201: Would this work, too?

```suggestion
            if isinstance(item['data'], (bytes, str)):
```

jpmckinney (commented):
I left one review comment. We can create a separate issue for my other comment if we want to address those separately.

yolile:
@jpmckinney I was trying to at least resolve the mexico_quien_es_quien issue but it seems like it will require more changes than expected so it is better to just move your comments to another issue (#616 )

jpmckinney (approved):

PR #613, opened 2021-02-17
read_data_middleware: close files

yolile:
ref #606

### DIFF
--- a/kingfisher_scrapy/middlewares.py
+++ b/kingfisher_scrapy/middlewares.py
@@ -273,7 +273,6 @@ class ReadDataMiddleware:
                 yield item
                 continue
             data = item['data'].read()
-            if hasattr(item['data'], 'close'):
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/middlewares.py
```diff
@@ -272,5 +272,8 @@ def process_spider_output(self, response, result, spider):
             if not isinstance(item, File) or not hasattr(item['data'], 'read'):
                 yield item
                 continue
-            item['data'] = item['data'].read()
+            data = item['data'].read()
+            if hasattr(item['data'], 'close'):
+                item['data'].close()
```
jpmckinney, line 277: I think the earlier check for a `read` attr would imply a `close` attr in our scenario, so we can remove the condition.

jpmckinney (approved):

PR #614, opened 2021-02-17
pluck: Remove some code for removed data types

jpmckinney:
While looking at the use of str/bytes/json.load/json.dump, I noticed the Pluck command still referenced old data types. I think some more cleanup can be done. I only did the easy and obvious so far.


jpmckinney:
We might be able to remove the `_get_*` methods entirely, since item data is now normalized.

yolile:
@jpmckinney yes, I was just about to create an issue for this, I will review and complete this

jpmckinney:
I'm plucking the license field, and had to make these changes to get that working.

Issue #523, opened 2020-10-16
IndexSpider: Code review


jpmckinney:
https://github.com/open-contracting/kingfisher-collect/pull/497 is great – I just noticed a few things.

* [x] Doesn't `_set_base_url` need to take kwargs that refer to `param_page`, `param_limit`, `param_offset`, rather than `page`, `limit` and `offset`? (**Update:** Actually, I don't think we need this method at all)
  * **Update**: Actually, I don't think we need this method at all: 8e1b0b9e033b1d747edf0815f69b91d58d6e5289
* [x] How do `limit_offset_range_generator` and `limit_offset_url_builder` behave if `self.limit` is set in the spider to a JSON pointer?
  * **Update**: They break. Added a test and fixed in 582b663cf860a60c75c7d2e875af714b14644d28
* [x] Instead of lines like `self.param_page = getattr(self, 'param_page', 'page')`, why not set them as class variables?
  * **Update**: No reason. Changed in 03add12af1f20a1391b4d53ed28295fe1531eb43
* [x] Should we set a default formatter of `self.formatter = staticmethod(parameters(self.param_page))` for `total_pages_pointer` or `self.formatter = staticmethod(parameters(self.param_offset))` for `count_pointer`?
  * **Update**: Why not. Done in 64adb100cb9ea3e9a8a7af098a33aef470ab598f
* Why does `page_size_range_generator` add 1 to the end of the range? I don't think any of the original spiders did that, and I don't think any of the spiders that were converted to use the IndexSpider strategy need it either.
  * **Update**: Nevermind. Simple cases like `list(range(2, ceil(11 / 10) + 1))` show why it's needed.
* [x] Instead of a try/except in `parse_list` (which can hide errors), there should be a class variable like `parse_list_loader` which defaults to `json.loads` (without parentheses), but that can be set to (for example) `int` for `kenya_makueni`.
  * **Update**: Fixed in 3f5e01f20b2f392741c6b4e7a928477f7e3a9e0e
* [x] In mexico_quien_es_quien.py, I think the new code will never download the first 1000 results. For `limit_offset_range_generator`, it might be better to use a new `start_index` instance variable instead of `self.limit`. Similarly, `page_size_range_generator` assumes that the spider wants to continue from page 2. If pages are 0-indexed, or if `start_requests` doesn't return data like in mexico_quien_es_quien.py, then a different page number / `start_index` would be apprioriate. We can set `start_index` to a reasonable default in the initializer.
[...]

PR #615, opened 2021-02-17
tests: Update some tests using old data types

jpmckinney:
This is complementary to #614. I wasn't sure if `test_index_spider.py` needed other updates. For example, one of the test cases uses `page_size`, but we don't define that in `base_spider`. Was `limit` intended?


yolile:
@jpmckinney I removed `page_size` from the test. We can explore what else needs to be removed in #523

yolile (approved):

jpmckinney:
Thanks for the new `join` test :)

PR #621, opened 2021-02-23
Big files spider

### LINKED ISSUE #620: Allow resize packages in non compressed file spiders too
`france` downloads a 693MB JSON file that kingfisher process is not able to process at the moment  (due to `["MemoryError()"]`)

we can add a `BigFile` base class that sets `resize_package = True` and overrides `parse` to:

``` python
    @handle_http_error
    def parse(self, response):
        data = {'data': response.body,
                'package': response.body}
        yield self.build_file(file_name=response.request.meta['file_name'], url=response.request.url,
                              data_type=self.data_type, data=data)
```

yolile:
closes #620


jpmckinney (approved):
Hopefully we won't need it for long :)

Issue #312, opened 2020-02-18
Explore options for streaming downloads in Scrapy


jpmckinney:
Presently, if a publisher only offers bulk downloads, and if we only want to download a sample, then we still need to download the entire bulk download, which can be large, like in the case of Digiwhist's (opentender.eu's) files, for example.

It's possible to gunzip a partial file (and I think also possible to untar a partial file). The only question is whether we can either: (1) request Scrapy to stop downloading after a given number of bytes or (2) stream the response to a callback that can then close the connection once it's read enough bytes.

jpmckinney:
Relevant to downloading large archive files like Digiwhist, which I think are presently loaded into memory.

jpmckinney:
> The only question is whether we can either: (1) request Scrapy to stop downloading after a given number of bytes or (2) stream the response to a callback that can then close the connection once it's read enough bytes.

[Scrapy 2.2](https://docs.scrapy.org/en/latest/news.html#scrapy-2-2-0-2020-06-24) adds a way to stop a download: 
https://docs.scrapy.org/en/latest/topics/request-response.html#topics-stop-response-download

jpmckinney:
When running `pluck`, I notice the memory usage grows and shrinks a fair bit. I assume this is to read the compressed files. In #629, I implement a solution to limit the bytes downloaded of non-compressed files.

jpmckinney:
With respect to closing this issue, we can consider allowing max-bytes to be specified in the `crawlall` command (which is designed for downloading samples), and maybe also in the crawl command (as a spider argument).

yolile:
> (2) large ZIPs, like georgia_opendata

could we use `resize_package` for these cases? (I haven't downloaded the file yet so I don't know if it is just one big JSON file or a lot of small ones)

jpmckinney:
I think the challenge is that to read even a part of a ZIP file, the ZIP file must be fully downloaded. Since we hold the ZIP file in memory (to avoid writing to disk which is a blocking operation), memory consumption will always be high for big ZIPs. It's not a problem so far, since the ZIPs are never too big.

PR #629, opened 2021-02-24
pluck: Add KINGFISHER_PLUCK_MAX_BYTES option to limit downloaded bytes

jpmckinney:
This adds a `--max-bytes` argument to the pluck command, so that we don't download huge files unnecessarily. (For now, we still need to download the huge compressed files, though maybe we can solve that later.)

There are a few publications that put their metadata in a "final" position, instead of before the releases/records. When plucking licenses, I noticed this for:

* moldova_old
* nepal_dhangadhi
* scotland_public_contracts

For these, the message in the pluck CSV is "error: /license not found within initial bytes", so it's clear to the user that maybe it appears later.

For now, I've made the behavior opt-in. Alternatively, we could add a spider class attribute to indicate whether package metadata appears in a final position, in which case it will read the full input even if `--max-bytes` is set.

With this option, the following command finishes relatively quickly:

```
scrapy pluck --logfile pluck.log --loglevel=WARN --max-bytes 10000 --package-pointer /license
```

Related: #312

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -13,6 +13,8 @@ Plucks one data value per publisher. It writes a CSV file with the results, and 
 -  ``--package-pointer=STR`` (``-p``): The JSON Pointer to the value in the package.
 -  ``--release-pointer=STR`` (``-r``): The JSON Pointer to the value in the release.
 -  ``--truncate=NUM`` (``-t``): Truncate the value to this number of characters.
+-  ``--max-bytes=NUM``: Stop downloading an OCDS file after reading at least this many bytes.
+-  ``spider``: Run specific spiders. Omit to run all spiders.
 
 Get each publisher's publication policy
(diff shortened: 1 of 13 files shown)


jpmckinney:
I haven't added tests yet, mainly because I don't yet know/haven't looked into how to test this :)

That said, plucking from Digiwhist with `--max-bytes` is now nearly instantaneous!

jpmckinney:
I’ll change it to max-chunks because that’s the actual Scrapy behavior (chunks are typically 16kb or less).

jpmckinney:
Okay, I'm done for now. I think ZIP and RAR files need to be downloaded in full and can't be streamed. 

If you have some ideas for testing, please feel free to add.

yolile:
@jpmckinney I have added some tests, let me know if you are happy with them

yolile (dismissed):

jpmckinney:
The tests look great, thanks!

PR #639, opened 2021-02-26
docs: add initial localization

yolile:


yolile:
@jpmckinney let me know if this is the preferred approach for localization, I don't think that we need a Transifex project for this

yolile:
I think that we can at least translate index.rst and some parts of spiders.rst

jpmckinney:
It's fine to translate in files directly. Please add a documentation page under `contributing/` about the translation setup. 

When ready, we'll have to do this setup to get `/es/latest/` to work: https://docs.readthedocs.io/en/stable/localization.html

yolile:
Actually, I realized now that we will need to have a separate project for the Spanish translation as https://docs.readthedocs.io/en/stable/localization.html#project-with-multiple-translations

jpmckinney:
Hmm, are you sure? We need a new ReadTheDocs project, but we can just specify the language in which to build the docs in ReadTheDocs settings.

PR #642, opened 2021-02-26
extensions: KingfisherPluck: Extensions are instantiated per crawler, not globally

jpmckinney:
I think the logic of checking which spiders have yielded items was introduced in https://github.com/open-contracting/kingfisher-collect/commit/e2e999d00ca5b936071482ae8a303a9037ae183f

### DIFF
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -20,10 +20,10 @@ class KingfisherPluck:
         self.directory = directory
         self.max_bytes = max_bytes
 
-        # The count of bytes received per spider.
-        self.bytes_received_counts = defaultdict(int)
-        # The set of spiders that have called the `item_scraped` method.
-        self.item_scraped_called = set()
+        # The number of bytes received.
+        self.total_bytes_received = 0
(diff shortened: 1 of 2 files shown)


jpmckinney:
https://docs.scrapy.org/en/latest/topics/extensions.html says

> Extensions are just regular classes that are instantiated at Scrapy startup, when extensions are initialized.

> Extensions are loaded and activated at startup by instantiating a single instance of the extension class.

But later:

> The main entry point for a Scrapy extension (this also includes middlewares and pipelines) is the from_crawler class method which receives a Crawler instance.

The latter is the correct expression, and it means there is one extension per crawler (and there is one crawler per spider).

As an example, the LogStats extension works in `crawlall` even though that extension doesn't track stats per spider: https://docs.scrapy.org/en/latest/_modules/scrapy/extensions/logstats.html#LogStats

I created a Scrapy PR to fix their documentation: https://github.com/scrapy/scrapy/pull/5014

yolile:
@jpmckinney yes, the documentation is kind of confusing, I think that is why I checked for each spider as well

yolile (approved):

PR #644, opened 2021-02-27
nepal_dhangadhi: Don't follow redirect to https://admin.ims.susasan.org/login

jpmckinney:


yolile:
@jpmckinney I've added one change now as we were missing one file with the current approach

jpmckinney:
I'm not clear what the `parse` method is doing?

The problem I'm trying to solve is that https://admin.ims.susasan.org/ocds/json/dhangadhi-2077-78.json redirects to https://admin.ims.susasan.org/login and will never succeed.


```
curl -I https://admin.ims.susasan.org/ocds/json/dhangadhi-2077-78.json
```

If I understand correctly, the new method will retry the 302, but allow the redirect, which will be to a bad page.

yolile:
> and will never succeed.

Actually, it only redirects the first time, but then, if we try again, it works! (and doesn't redirect anymore either)

jpmckinney:
Weird, I can run curl many times and it's always the same. I'll try the updated code.

jpmckinney:
Hmm, for me, `dhangadhi-2077-78.json` downloads as an HTML file, which we don't want.

yolile:
you are right :(  I added a Caveats though, feel free to edit it if you want

yolile (approved):

PR #646, opened 2021-02-27
crawlall: Optionally pass specific spiders as arguments

jpmckinney:


yolile (approved):

yolile:
@jpmckinney should we update https://kingfisher-collect.readthedocs.io/en/latest/cli.html#crawlall too?

jpmckinney:
> @jpmckinney should we update https://kingfisher-collect.readthedocs.io/en/latest/cli.html#crawlall too?

Good catch. I copied the same docs as for pluck.

yolile (approved):

PR #650, opened 2021-02-28
indonesia_bandung: Remove logic introduced in #140

jpmckinney:
I ran the spider locally, and I never encountered an empty response. I'm not sure in what condition the original code was needed.


yolile:
> I ran the spider locally, and I never encountered an empty response. I'm not sure in what condition the original code was needed.

Probably an empty response could arrive in the past but not anymore

yolile (approved):

PR #657, opened 2021-03-04
AddPackageMiddleware: add spider declared version to basic package

yolile:
I think that we can add the spiders's declared version to the basic package as well so we don't need to transform the packages from 1.0 to 1.1 in Kingfisher Process if the releases are already in the latest version of the standard


jpmckinney (approved):
Nice catch!

PR #676, opened 2021-03-18
673 indonesia

### LINKED ISSUE #673: Indonesia - opentender.net
@yolile Indonesia Corruption Watch just shared their new OCDS publication in CRM-6782: https://opentender.net/#/ocds

They can't provide a single bulk file because of restrictions in their agreement with the national procurement authority. However, I'm assuming the download page as implemented would be particularly difficult to scrape.

Should we ask them to also provide a page with a list of download URLs for each LPSE (procuring entity)?

cc @jpmckinney

cecicasco:
closes #673


kingfisher_scrapy/spiders/indonesia_opentender.py
```diff
@@ -0,0 +1,46 @@
...
+
+    base_url = 'https://opentender.net/api/'
+    # PeriodicSpider
+    pattern = base_url + 'master/lpse?year={}'
+    start_requests_callback = 'parse_list'
+
+    @handle_http_error
+    def parse_list(self, response):
+        data = response.json()
+        date = response.request.meta['date']
```
yolile, line 37: I think that instead of modifying the base calls we can do something like this: (get the year from the request URL)
```suggestion
        year = parameters('year')(response.request.url).split('-')[1]
```

kingfisher_scrapy/spiders/indonesia_opentender.py
```diff
@@ -0,0 +1,46 @@
...
+        for item in data['data']:
+            code = item['code']
```
yolile, line 39: I tested this and I got some request duplication warnings, which mean that we are having duplicates in the response, so we can do something like 
```python
        requested_codes = set()
        for item in data['data']:
            code = item['code']
            # there are some duplicated codes
            if code and code not in requested_codes:
                requested_codes.add(code)
               ....
```

yolile (changes requested):
@cecicasco Congratulations on coding your first spider! I just left a couple of comments. And also please don't forget to run `scrapy updatedocs` https://kingfisher-collect.readthedocs.io/en/latest/contributing/index.html#commit-the-spider

kingfisher_scrapy/spiders/indonesia_opentender.py
```diff
@@ -21,7 +21,7 @@ class IndonesiaOpentender(CompressedFileSpider, PeriodicSpider):
 
     # BaseSpider
     date_format = 'year'
-    default_from_date = '2008'
+    default_from_date = '2020'
```
yolile, line 24: @cecicasco should this be 2008?

kingfisher_scrapy/spiders/indonesia_opentender.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @handle_http_error
+    def parse_list(self, response):
+        data = response.json()
+        year = response.request.url.split('=')[1]
+        requested_codes = []
+        for item in data['data']:
+            code = item['code']
+            # there are some duplicated codes
+            if code not in requested_codes:
```
yolile, line 42: In the previous version, you checked if `code` is not `None`  so I assume that is still needed?
```suggestion
            if code and code not in requested_codes:
```

cecicasco: I gonna add again, you are right..

yolile (approved):

PR #683, opened 2021-03-19
682 validation pipeline

### LINKED ISSUE #682: Validation Pipeline: drop duplicated items instead of just warning
And update https://kingfisher-collect.readthedocs.io/en/latest/logs.html indicating that we should (and how) look for dropped items too

yolile:
closes #682

### DIFF
--- a/tests/pipelines/test_validate.py
+++ b/tests/pipelines/test_validate.py
@@ -105,6 +105,9 @@ def test_process_item_duplicate_file(caplog):
     pipeline.process_item(item, spider)
     with pytest.raises(DropItem) as excinfo:
         pipeline.process_item(item, spider)
+    item2 = item.copy()
+    item2['file_name'] = 'file2'
+    pipeline.process_item(item2, spider)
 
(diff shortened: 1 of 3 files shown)


tests/pipelines/test_validate.py
```diff
@@ -102,13 +103,8 @@ def test_process_item_duplicate_file(caplog):
...
 
     pipeline.process_item(item, spider)
-    pipeline.process_item(item, spider)
-    item2 = item.copy()
-    item2['file_name'] = 'file2'
-    pipeline.process_item(item2, spider)
-
-    assert len(caplog.messages) == 1
-    assert caplog.messages[0] == "Duplicate File: 'test1'"
+    with pytest.raises(DropItem):
```
jpmckinney, line 106: you can do with ... as excinfo and then test `str(excinfo.value)`

yolile: done and I've updated the message too!

jpmckinney:
I've added a couple commits that restore a part of the test and keep the message closer to before.

jpmckinney (approved):

PR #684, opened 2021-03-19
extensions: Twisted Failure's were not serializable (fix #680)

jpmckinney:
Now with tests to catch my earlier mistake :)


yolile:
I was going to ask you in the other PR if testing for these was very complicated, but I didn't want to be so intense!

jpmckinney:
Testing is always good, no worries :)

yolile (approved):

PR #686, opened 2021-03-25
Add skip pluck attribute to more spiders

### LINKED ISSUE #628: skip_pluck: Add class attribute to additional spiders
I added it to some in 4d05c23d245c8286849842c2ecb9dcbe09fdfe62, f0e741622c281123dc2ead7bbba2ce3aae88f696 and 8ba0c111e7f13564f585bbf7fe0b0b539ab78f1c

Can we add it to the following, or are they different?

* portugal (should maybe be renamed portugal_bulk)
* pakistan_ppra_bulk

To my knowledge, the only bulk download that is different from the corresponding API is `chile_compra_bulk`.

### LINKED ISSUE #647: pluck: Using it to determine latest release date won't work with IndexSpider, PeriodicSpider
We've made spider classes that are extremely efficient at scheduling requests. However, to determine the latest release date, the order of requests matters. Some options:

* Set from_date. For example, for MEL1 criteria, we can set it to the start of the year, since any date within the year is fine
* Stop yielding requests from `build_request` as soon as there is one without a callback

yolile:
closes #628
closes #647 

This also removes scotland_base as a base class is not necessary anymore as we have only one spider for scotland


kingfisher_scrapy/spiders/digiwhist_base.py
```diff
@@ -16,6 +16,7 @@ class DigiwhistBase(BaseSpider):
     """
     # BaseSpider
     line_delimited = True
+    skip_pluck = 'non ordered releases'
```
yolile, line 19: actually I'm not sure if we should skip them for not being in order as we can still use pluck for non date values

jpmckinney: Yes, let's leave it in. We should only `skip_pluck` for spiders that would provide the same answer.

kingfisher_scrapy/base_spider.py
```diff
@@ -169,7 +169,7 @@ def get_start_time(self, format):
             date = self.crawler.stats.get_value('start_time')
         return date.strftime(format)
 
-    def build_request(self, url, formatter, **kwargs):
+    def build_request(self, url, formatter, priority=0, **kwargs):
```
jpmckinney, line 172: Since we're just passing through the priority, we could leave this method as is (it will be passed in `kwargs`).

jpmckinney (commented):
For `range_generator`, I guess we assume that the API paginates in reverse chronological order? Is this true for all applicable spiders? If not, we can have a class attribute like `chronological_order = 'asc'`

kingfisher_scrapy/base_spider.py
```diff
@@ -592,7 +596,10 @@ def pages_url_builder(self, value, data, response):
     def limit_offset_range_generator(self, data, response):
         limit = self._resolve_limit(data)
         count = resolve_pointer(data, self.count_pointer)
-        return range(self.limit, count, limit)
+        if not self.yield_list_results:
+            return range(0, count, limit)
+        else:
+            return range(self.limit, count, limit)
```
jpmckinney, line 602: What is this fixing?

kingfisher_scrapy/spiders/mexico_administracion_publica_federal_api.py
```diff
@@ -16,6 +16,7 @@ class MexicoAdministracionPublicaFederalAPI(IndexSpider):
 
     # BaseSpider
     root_path = 'results.item'
+    skip_pluck = 'Already covered (see code for details)'  # mexico_administracion_publica_federal_bulk.py
```
jpmckinney, line 19: Remove .py (also in mexico_nuevo_leon_releases.py)

kingfisher_scrapy/spiders/zambia.py
```diff
@@ -26,7 +26,8 @@ def start_requests(self):
 
     @handle_http_error
     def parse_list(self, response):
-        urls = response.json()['packagesPerMonth']
+
+        urls = reversed(response.json()['packagesPerMonth'])
```
jpmckinney, line 30: I already fixed this in HEAD by reversing urls in the for loop.

kingfisher_scrapy/base_spider.py
yolile: the offset=0 is never downloaded otherwise, related to https://github.com/open-contracting/kingfisher-collect/issues/523 bullet 7

jpmckinney: Aha. To reduce duplication and explain the logic:

```suggestion
        # If `yield_list_results` is `True` (default), the response is parsed is `parse_list`.
        if self.yield_list_results:
            start = self.limit
        else:
            start = 0
        return range(start, count, limit)
```

yolile: ...and that is why I shouldn't work late and tired!

yolile:
> For `range_generator`, I guess we assume that the API paginates in reverse chronological order? Is this true for all applicable spiders? If not, we can have a class attribute like `chronological_order = 'asc'`

I've added `chronological_order` default to "desc", and checked all the IndexSpider classes and didn't find one with asc chronological order, they are all desc or non-ordered

jpmckinney (dismissed):

yolile:
@jpmckinney I assume that I can merge this as you approved it before your newest commit :)

Issue #600, opened 2021-01-21
New spider class?


aguilerapy:
After [PR 596](https://github.com/open-contracting/kingfisher-collect/pull/596), Afghanistan, Colombia, Costa Rica, Dominican Republic, Malta, Zambia (and others) will filter the dates from the URLs of their base lists.

Should we create a new spider class? They all have a list of available URLs in common, and they can have a method like this:

```
    def parse_dates_to_filter(self, date):
        if self.from_date and self.until_date:
            return False

        if self.date_format == '%Y':
            if not (self.from_date.year <= date <= self.until_date.year):
                return True
        elif self.date_format == '%Y-%m':
            return not ((self.from_date.year <= date.year <= self.until_date.year)
                        and (self.from_date.month <= date.month <= self.until_date.month))
        else:
            date = datetime.strptime(date, self.date_format)
            return not (self.from_date <= date <= self.until_date)
```

We have the `PeriodicSpider` class but it doesn't work with those scrapers.

jpmckinney:
Sounds good to me! @yolile ?

yolile:
Yes, the new class sounds good to me as well, although instead of having the suggested method, the class could have the `date_format` and `date_pattern` attribute, `start_request` method with a callback to a `build_urls` function where the returned URL list is filtered and yielded according to the `date_format` and `date_pattern`

jpmckinney:
See discussion in https://github.com/open-contracting/kingfisher-collect/pull/701#discussion_r617035196 about updating Afghanistan, Zambia and Malta.

jpmckinney:
* Afghanistan: gone
* Colombia: gone
* Costa Rica: CKANSpider
* Malta: EuropeanDynamicsBase
* Zambia: EuropeanDynamicsBase

Leaving maybe Dominican Republic "(and others)", but I'm not sure what logic this issue is meant to extract (perhaps we did it independently, already), so closing this issue.

PR #701, opened 2021-04-20
moldova: add from_date support

yolile:
ref #346

### DIFF
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -57,6 +57,7 @@ class BaseSpider(scrapy.Spider):
     line_delimited = False
     root_path = ''
     dont_truncate = False
+    default_from_date = None
 
     def __init__(self, sample=None, note=None, from_date=None, until_date=None, crawl_time=None,
                  keep_collection_open=None, package_pointer=None, release_pointer=None, truncate=None, *args,
--- a/kingfisher_scrapy/base_spider.py
+++ b/kingfisher_scrapy/base_spider.py
@@ -57,7 +57,6 @@ class BaseSpider(scrapy.Spider):
     line_delimited = False
     root_path = ''
     dont_truncate = False
-    default_from_date = None
 
     def __init__(self, sample=None, note=None, from_date=None, until_date=None, crawl_time=None,
                  keep_collection_open=None, package_pointer=None, release_pointer=None, truncate=None, *args,
--- a/kingfisher_scrapy/spiders/moldova.py
+++ b/kingfisher_scrapy/spiders/moldova.py
@@ -19,6 +19,8 @@ class Moldova(SimpleSpider):
 
     # BaseSpider
     date_format = 'datetime'
+    default_from_date = '2018-10-18T00:00:00'
+    date_required = True
 
     def start_requests(self):
         # https://public.mtender.gov.md offers three endpoints: /tenders/, /tenders/plan/ and /budgets/. However, this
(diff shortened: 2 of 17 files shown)


kingfisher_scrapy/base_spider.py
```diff
@@ -57,6 +57,7 @@ class attribute to the path to the OCDS data.
     line_delimited = False
     root_path = ''
     dont_truncate = False
+    default_from_date = None
```
jpmckinney, line 60: Why add this default? I figure we'd prefer an `AttributeError` if a spider forgot to set the attribute?

kingfisher_scrapy/spiders/moldova.py
```diff
@@ -8,12 +8,20 @@ class Moldova(SimpleSpider):
...
+        Download only data from this time onward (YYYY-MM-DDThh:mm:ss format).
     """
     name = 'moldova'
 
     # SimpleSpider
     data_type = 'release_package'
 
+    # BaseSpider
+    date_format = 'datetime'
+    default_from_date = '2018-10-18T00:00:00'
```
jpmckinney, line 22: I suggest 2018-01-01 (we can even do 2010-01-01), since we don't need to have the exact from date for this API. We should only set a precise from date if the API will error otherwise. That way, if the API fills in historical data, we don't accidentally omit it.

I think the afghanistan spiders are also too precise.

There are some that have a precise month - I don't know if they can be relaxed as well: honduras_portal_bulk, malta, uruguay_base, zambia.

We can update the docstring about setting `default_from_date`.

yolile: Actually, I didn't want to add a default_date for this one, as it is not required, but https://github.com/open-contracting/kingfisher-collect/blob/81e9a22d676a95c01d0aaedf0d2a92f3601173a2/kingfisher_scrapy/commands/checkall.py#L121 fails otherwise.

jpmckinney: That check is there, because otherwise how will the f-string be formatted?

yolile: The default_from_date is used in:
https://github.com/open-contracting/kingfisher-collect/blob/81e9a22d676a95c01d0aaedf0d2a92f3601173a2/kingfisher_scrapy/commands/checkall.py#L160-L165
https://github.com/open-contracting/kingfisher-collect/blob/81e9a22d676a95c01d0aaedf0d2a92f3601173a2/kingfisher_scrapy/commands/checkall.py#L182

but I think that if `date_required` is false and only `from_date` or only `until_date` is implemented, then we don't need a `default_from_date` or a `default_until_date`, right?

jpmckinney: Yes, it's not required in that specific scenario.

yolile: So maybe I can update this check and remove `default_from_date` from Moldova?

jpmckinney: Sounds good, but I think the f-string also needs to be updated as right now it would format it with `None`, no?

yolile: > Sounds good, but I think the f-string also needs to be updated as right now it would format it with None, no?
Done, and I also added that the check ask if there is a default_until_date
I also updated all the warnings that we had

> I think the afghanistan spiders are also too precise.
There are some that have a precise month - I don't know if they can be relaxed as well: honduras_portal_bulk, malta, uruguay_base, zambia.

Honduras and Uruguay are fine as they have fixed dates that are not going to change. For Afghanistan, zambia and Malta,  the spider ask for both dates to be set (from_date and until_date) and I think that is not required, but I will update that as part of #600

yolile:
@jpmckinney now the coverage is failing because of `checkall` but I think that is ok

jpmckinney (approved):
I made some changes. Please have a look and feel free to merge.

jpmckinney:
> @jpmckinney now the coverage is failing because of `checkall` but I think that is ok

That's fine, it is not a required check.

PR #707, opened 2021-04-22
commands: add incremental data store

yolile:


kingfisher_scrapy/commands/incremental_data_store/db.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+import psycopg2 as psycopg2
+
+
+# from https://github.com/open-contracting/kingfisher-summarize/blob/main/ocdskingfishersummarize/db.py
+class Database:
```
jpmckinney, line 7: Instead of a long Database class, since most methods are only called from one location, can we in-line most of this into the command itself?

kingfisher_scrapy/commands/incremental_data_store/incrementaldatastore.py
```diff
@@ -0,0 +1,86 @@
...
+    def run(self, args, opts):
+        spiders = self.crawler_process.spider_loader.list()
+        spider_name = opts.spider
+
+        if not spider_name or opts.spider not in spiders:
+            raise UsageError('A valid spider must be given.')
+
+        # we disable kingfisher process extension
+        extensions = {'kingfisher_scrapy.extensions.KingfisherProcessAPI': None}
+        self.settings.set('EXTENSIONS', extensions)
```
jpmckinney, line 68: Won't this disable all extensions?

kingfisher_scrapy/commands/incremental_data_store/db.py
yolile: sure, I found having the separate class nicer to read, but you are right

kingfisher_scrapy/commands/incremental_data_store/incrementaldatastore.py
yolile: true, fixed

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,126 @@
...
+        cursor.execute(f'SET search_path = {SCHEMA_NAME}')
+        self.create_table(cursor, spider_name)
+        return connection, cursor
+
+    def create_table(self, cursor, spider_name):
+        """
+        Creates a table with a jsonb "data" column and creates and index on the data->>'date' field
+        """
+        cursor.execute(f'CREATE TABLE IF NOT EXISTS {spider_name} (data jsonb);')
+        cursor.execute(f"CREATE INDEX IF NOT EXISTS idx_{spider_name} ON {spider_name}(cast(data->>'date' as text));")
```
jpmckinney, line 61: It's a bit faster to create an index after all the data is inserted.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def json_to_csv(self, data_directory):
+        """
+        Reads all the OCDS release packages json files from the given directory and save each release as a row in a
+        csv file called data.csv
+        """
+        file_name = os.path.join(data_directory, 'data.csv')
+        with open(file_name, 'w') as csv_file:
+            csv_writer = csv.writer(csv_file)
+            for filename in os.listdir(data_directory):
```
jpmckinney, line 85: `scandir` is faster on directories with many files: https://docs.python.org/3/library/os.html#os.scandir

But, I think this will only be relevant if `--compile` isn't set.

kingfisher_scrapy/commands/incrementaldatastore.py
yolile: changed

yolile:
@jpmckinney not sure why the test is falling as the `psycopg2` is declared in the requirements files

jpmckinney:
@yolile I think you need to install the libpq-dev Ubuntu package for psycopg2 to install.

yolile:
> @yolile I think you need to install the libpq-dev Ubuntu package for psycopg2 to install.

I just forgot to commit the requirements_dev.txt file. But now I'm having a new error when running the test_requirements test: `AssertionError: Missing requirements for modules: ocdsmerge`

jpmckinney:
> > @yolile I think you need to install the libpq-dev Ubuntu package for psycopg2 to install.
> 
> I just forgot to commit the requirements_dev.txt file. But now I'm having a new error when running the test_requirements test: `AssertionError: Missing requirements for modules: ocdsmerge`

You removed ocdsmerge from requirements.in (and others) but it's used by pipelines.py, so the test fails.

Once tests pass, let me know and I'll review.

yolile:
@jpmckinney oh true! sorry, now the tests pass

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        cursor = connection.cursor()
+        cursor.execute(f'SET search_path = {schema_name}')
+        self.create_table(cursor, spider_name)
+        return connection, cursor
+
+    def create_table(self, cursor, spider_name):
+        """
+        Creates a table with a jsonb "data" column and creates and index on the data->>'date' field
+        """
+        cursor.execute(f'CREATE TABLE IF NOT EXISTS {spider_name} (data jsonb);')
```
jpmckinney, line 63: We can omit semi-colons from all SQL statements.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # gets the last data->>'date' that exists in the country table
+        cursor.execute(f"SELECT max(data->>'date') FROM {spider_name};")
+        last_date = cursor.fetchone()[0]
+
+        logger.info(f'Last release date found {last_date}')
+
+        self.run_spider(last_date, spidercls)
+
+        # we drop the table and insert all the data again
```
jpmckinney, line 131:
```suggestion
        # Replace the spider's data table.
```

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+            csv_file_name = self.json_to_csv(merge(self.get_data_from_directory(data_directory)),
+                                             data_directory)
+        # if we don't need to compile the releases, we insert them directly in the database
+        else:
+            if 'release' in spidercls.data_type:
+                list_type = 'releases.item'
+            else:
+                list_type = 'records.item.compiledRelease'
+            csv_file_name = self.json_to_csv(self.get_data_from_directory(data_directory, list_type),
+                                             data_directory)
```
jpmckinney, line 148: We can eliminate some repetition here, to only call `self.get_data_from_directory` and `self.json_to_csv` once. If we do that, we can actually in-line `json_to_csv`.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        cursor.execute(f"SELECT max(data->>'date') FROM {spider_name};")
+        last_date = cursor.fetchone()[0]
+
+        logger.info(f'Last release date found {last_date}')
+
+        self.run_spider(last_date, spidercls)
+
+        # we drop the table and insert all the data again
+        cursor.execute(f'DROP TABLE {spider_name} CASCADE ;')
+        self.create_table(cursor, spider_name)
```
jpmckinney, line 133: Let's move this right before the insertion so that the DB logic is next to each other.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            if 'release' in spidercls.data_type:
+                list_type = 'releases.item'
+            else:
+                list_type = 'records.item.compiledRelease'
+            csv_file_name = self.json_to_csv(self.get_data_from_directory(data_directory, list_type),
+                                             data_directory)
+        logger.info('Dumping the data into the data base')
+        with open(csv_file_name) as csv_file:
+            cursor.copy_expert(f"COPY {spider_name}(data) FROM STDIN WITH CSV", csv_file)
```
jpmckinney, line 151: I prefer just `f` for file pointers:

```suggestion
        with open(csv_file_name) as f:
            cursor.copy_expert(f"COPY {spider_name}(data) FROM STDIN WITH CSV", f)
```

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        os.remove(csv_file_name)
+        cursor.execute(f"CREATE INDEX IF NOT EXISTS idx_{spider_name} ON {spider_name}(cast(data->>'date' as text));")
+        connection.commit()
+
+        logger.info('Process completed')
```
jpmckinney, line 156: Do we need this message?

```suggestion
```

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        # we disable kingfisher process extension
+        self.settings['EXTENSIONS']['kingfisher_scrapy.extensions.KingfisherProcessAPI'] = None
+
+        connection, cursor = self.database_setup(spider_name, opts.db_schema)
+
+        # gets the last data->>'date' that exists in the country table
+        cursor.execute(f"SELECT max(data->>'date') FROM {spider_name};")
+        last_date = cursor.fetchone()[0]
+
+        logger.info(f'Last release date found {last_date}')
```
jpmckinney, line 127: For clarity to the user, I prefer to log the command that we're simulating, e.g. 

```suggestion
        logger.info(f'Running: scrapy crawl -a from_date={kwargs["from_date"]} -a crawl_time={kwargs["crawl_time"]}')
```

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        self.create_table(cursor, spider_name)
+        return connection, cursor
+
+    def create_table(self, cursor, spider_name):
+        """
+        Creates a table with a jsonb "data" column and creates and index on the data->>'date' field
+        """
+        cursor.execute(f'CREATE TABLE IF NOT EXISTS {spider_name} (data jsonb);')
+
+    def run_spider(self, last_date, spidercls):
```
jpmckinney, line 65: Since this is a short method that is only called from one location, let's put it in `run()` so that the developer doesn't need to scroll as much.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        kwargs = {'crawl_time': CRAWL_TIME}
+
+        # If there is data already in the database we only download data after the last release date
+        if last_date:
+            kwargs['from_date'] = self.from_date_formatted(last_date, spidercls.date_format)
+        self.crawler_process.crawl(spidercls, **kwargs)
+        self.crawler_process.start()
+
+    def get_data_from_directory(self, data_directory, json_data_path=''):
```
jpmckinney, line 74: ijson uses the term `prefix` instead of `json_data_path`.

kingfisher_scrapy/commands/incrementaldatastore.py
```diff
@@ -0,0 +1,156 @@
...
+        Receives an iterable of json dicts and save them into a data.csv file in the given directory
+        :param data an iterable of json dict
+        :param data_directory directory to where create the csv file
+        """
+        file_name = os.path.join(data_directory, 'data.csv')
+        with open(file_name, 'w') as csv_file:
+            csv_writer = csv.writer(csv_file)
+            for item in data:
+                data = json.dumps(item, default=util.default)
+                csv_writer.writerow([data])
```
jpmckinney, line 96:
```suggestion
        with open(file_name, 'w') as f:
            writer = csv.writer(f)
            for item in data:
                data = json.dumps(item, default=util.default)
                writer.writerow([data])
```

jpmckinney:
Deployment notes:

- Need libpq-dev for psycopg2 to compile
- Set KINGFISHER_COLLECT_DATABASE_URL environment variable when adding cron job

jpmckinney:
> Use the first word of the spider's name as the table name to avoid table names like 'chile_compra_records'

What is wrong with `chile_compra_records`? Splitting on `_` means we get tables like `dominican` or `costa`.

yolile:
> > Use the first word of the spider's name as the table name to avoid table names like 'chile_compra_records'
> 
> What is wrong with `chile_compra_records`? Splitting on `_` means we get tables like `dominican` or `costa`.

If you are ok with "chile_compra_records" and similar I can leave it like that. The other option is to ask for the table name in the command line

jpmckinney:
I think it is clearer to reference the spider. I'm making a few small changes, so I can commit that change.

jpmckinney:
> @jpmckinney let me know if I should add some tests for the new command

We should add some tests. We can maybe add a `success.py` spider like `fail.py` which yields a few items.

if so, we'll need to ignore it in crawlall, pluck and updatedocs.

jpmckinney:
It looks like record spiders are only supported if they set `compiledRelease`. I'll just add it to the docs for now.

jpmckinney:
Do we need `CASCADE`? I don't expect any other tables to depend on these tables.

I've made my commit.

yolile:
> Do we need CASCADE? I don't expect any other tables to depend on these tables.

It was only because of the index, but I see that the DROP sentence actually drop the existing indexes already, so I will remove CASCADE

jpmckinney:
Why is `w+` needed? I don't see where we are reading from the file, in which case `w` is enough.

yolile:
@jpmckinney this is ready for you to review again

jpmckinney:
Deployment issue is https://github.com/open-contracting/kingfisher-collect/pull/707

jpmckinney (dismissed):
I committed my one suggestion. I did not know about `capsys`! I've updated other libraries to use it instead of patching `sys.stdout` and `sys.stderr`.

jpmckinney (dismissed):

jpmckinney (approved):

jpmckinney:
I'm not 100% sure about the name `incrementaldatastore` but I can't think of a better name (`crawltodb`, etc.).

yolile:
@jpmckinney I'm not sure either! Actually, the "incremental" part is only true if you use the command regularly, if you use it just once then is just `crawltodb` as you mentioned.

jpmckinney (commented):
Now that I think about it - should this be an extension with most of the logic in `spider_closed`? It can be activated when a `database_url` spider argument is set.

We'd need to be able to set `from_date` in `spider_opened`.

yolile:
> Now that I think about it - should this be an extension with most of the logic in spider_closed? It can be activated when a database_url spider argument is set.

I remember that we discussed that option, but I'm not sure why we discarded that idea. If we do that, we will need to deactivate Kingfisher Process too.

jpmckinney:
Yeah, not sure how to disable `KingfisherProcessAPI` from another extension... I guess it has to be a command!

yolile:
> We'd need to be able to set from_date in spider_opened.

And the option will need to be used with crawl_time

jpmckinney:
Actually, if we make the `DATABASE_URL` a Scrapy setting instead (set with `-s DATABASE_URL=...`) then it is visible to extensions (whereas spider arguments are not), so it's possible to make this an extension. Hmm... enough for tonight!

yolile:
closing in favor of #719

PR #711, opened 2021-04-29
Fix australia spider

### LINKED ISSUE #399: australia_nsw: Check pagination logic
For the tender and contract search endpoints, it seems to return a next link, even if the current page contains no data (you can set an arbitrarily large `startRow` and it will return a response with a new `next` link).

nativaldezt:
Closes #399


nativaldezt:
Should I add a comment like "When there are no more releases there is still a link to next page so it is required to check data['releases']" ?

yolile (approved):

PR #719, opened 2021-05-21
extensions: add DataBaseStore

yolile:


yolile:
@jpmckinney this is how the command as an extension looks like. The only "bad" things that I found was:
- We need to use two separate database connections, one for creating the table and querying the last date in `spider_opened`, and another one in `spider_closed` for replacing the table
- We cannot close or finish a spider from the `spider_opened` method, so if the `crawl_time` is not set, or if `compile` is used with records or if `sample` is set, we can just raise an exception or warning but not actually stopping the execution.

jpmckinney:
* Agreed on call that two connections is fine, especially since they might be far apart and the connection might not live that long.
* Discussed on call trying the CloseSpider exception and/or raising the error from the spider instead of from the extension (which reduces encapsulation, but that's okay).

kingfisher_scrapy/base_spider.py
```diff
@@ -80,6 +81,9 @@ def __init__(self, sample=None, note=None, from_date=None, until_date=None, craw
         self.release_pointer = release_pointer
         self.truncate = int(truncate) if truncate else None
 
+        # DatabaseStore-related argument
+        self.compile = compile_releases == 'true'
```
jpmckinney, line 85: Should we name it `compile_releases` for consistency?

kingfisher_scrapy/base_spider.py
```diff
@@ -80,6 +81,9 @@ def __init__(self, sample=None, note=None, from_date=None, until_date=None, craw
         self.release_pointer = release_pointer
         self.truncate = int(truncate) if truncate else None
 
+        # DatabaseStore-related argument
```
jpmckinney, line 84:
```suggestion
        # DatabaseStore-related argument.
```

kingfisher_scrapy/base_spider.py
```diff
@@ -107,6 +111,7 @@ def __init__(self, sample=None, note=None, from_date=None, until_date=None, craw
             'package_pointer': package_pointer,
             'release_pointer': release_pointer,
             'truncate': truncate,
+            'compile': compile_releases
```
jpmckinney, line 114: Always put trailing commas, so that when we add another line, we don't have to edit this line (smaller diff).

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
                 json.dump(data, f, default=util.default)
 
 
+class DatabaseStore:
+    """
+    If DATABASE_URL is set and crawl_time used, store the spider data in a PostgresSQL database, incrementally
```
jpmckinney, line 144:
```suggestion
    If the ``DATABASE_URL`` Scrapy setting and the ``crawl_time`` spider argument are set, store the spider data in a
    PostgresSQL database, incrementally.
```

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+
+    A table with a \"data\" column is created if it doesn't exist, named after the spider. If the table isn't empty,
+    the crawl starts with the `from_date` spider argument set to the maximum value of the `date` field of the OCDS data
+    stored in the \"data\" column. If the spider returns records, each record must set the `compiledRelease` field."
```
jpmckinney, line 148:
```suggestion
    A table with a "data" column is created if it doesn't exist, named after the spider. If the table isn't empty, the
    crawl starts with the ``from_date`` spider argument set to the maximum value of the ``date`` field of the OCDS data
    stored in the "data" column. If the spider returns records, each record must set the ``compiledRelease`` field.
```

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+    """
+
+    connection = None
+    cursor = None
+    crawl_directory = None
+    file_store_directory = None
```
jpmckinney, line 154: Since the extension is `FilesStore`, this should be `files_store_directory`

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+            raise NotConfigured('FILES_STORE is not set.')
+
+        extension = cls(database_url, directory)
+
+        crawler.signals.connect(extension.spider_opened, signal=signals.spider_opened)
+        crawler.signals.connect(extension.spider_closed, signal=signals.spider_closed)
+
+        return extension
+
+    def format_from_date(self, date, date_format, valid_formats):
```
jpmckinney, line 181: Is there a way to reuse the logic from the spider? Can we parse the date from the DB, and then call `datetime.strptime(from_date, spider.date_format)` like in BaseSpider?

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+        if date_format == valid_formats['datetime']:
+            return date[:19]
+        if date_format == valid_formats['date']:
+            return date[:10]
+        if date_format == valid_formats['year-month']:
+            return date[:7]
+        return date[:4]
+
+    def spider_opened(self, spider):
+
```
jpmckinney, line 191: Avoid blank line after method signature.

```suggestion
```

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+        if not spider.crawl_time:
+            self.data_use_error = 'The crawl_time argument must be set'
+
+        if spider.compile and 'record' in spider.data_type:
+            self.data_use_error = 'The compile flag can only be set if the spider returns releases.'
+
+        # We raise an error but the spider is not closed or finished. There is no way to close or finish the spider
+        # from this method.
+        if self.data_use_error:
```
jpmckinney, line 200: As discussed on call, hopefully we can avoid `self.data_use_error`.

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+            self.logger.info('Getting the date from which to resume the crawl (if any)')
+            self.execute("SELECT max(data->>'date') FROM {table}", table=spider.name)
+            from_date = self.cursor.fetchone()[0]
+            if from_date:
+                spider.from_date = datetime.datetime.strptime(self.format_from_date(from_date, spider.date_format,
+                                                                                    spider.VALID_DATE_FORMATS),
+                                                              spider.date_format)
+        self.connection.commit()
+        self.cursor.close()
+        self.connection.close()
```
jpmckinney, line 220: This should be in a `finally:`, with the `try:` starting after the cursor is assigned.

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+        filename = os.path.join(crawl_directory_full_path, 'data.csv')
+        with open(filename, 'w') as f:
+            writer = csv.writer(f)
+            for item in data:
+                writer.writerow([json.dumps(item, default=util.default)])
+
+        self.logger.info('Replacing the JSON data in the SQL table')
+        try:
+            self.connection = psycopg2.connect(self.database_url)
+            self.cursor = self.connection.cursor()
```
jpmckinney, line 254: I think these need to be outside the `try`, otherwise if the connection fails, then the cursor is not set, and we'll get an exception in the `finally`.

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,166 @@ def _write_file(self, path, data):
...
+        return sql.SQL(statement).format(**objects)
+
+    def create_table(self, table):
+        self.execute('CREATE TABLE IF NOT EXISTS {table} (data jsonb)', table=table)
+
+    def yield_items_from_directory(self, data_directory, prefix=''):
+        for dir_entry in os.scandir(data_directory):
+            if dir_entry.name.endswith('.json'):
+                with open(dir_entry.path) as f:
+                    yield from ijson.items(f, prefix)
```
jpmckinney, line 290: Move these above `format`, so that the Summarize code is together.

kingfisher_scrapy/settings.py
```diff
@@ -122,6 +123,7 @@
 KINGFISHER_PLUCK_PATH = os.getenv('KINGFISHER_PLUCK_PATH', '')
 KINGFISHER_PLUCK_MAX_BYTES = None
 
+DATABASE_URL = os.getenv('KINGFISHER_COLLECT_DATABASE_URL')
```
jpmckinney, line 126: I think it might be better to set it to `None`, so that the only way to set it is with CLI arguments. Otherwise, we might accidentally set it and wonder why data isn't going to Process.

kingfisher_scrapy/extensions.py
yolile: I think that is what we are doing already, the date from the database is a full datetime date and we need to truncate it by datetime (no changes), date, year-month, or year, and then we use `datetime.strptime(from_date, spider.date_format)`

yolile:
> Discussed on call trying the CloseSpider exception and/or raising the error from the spider instead of from the extension (which reduces encapsulation, but that's okay).

The CloseSpider exception is raised but the spider is not closed:, so I will go ahead with the second option 
image

kingfisher_scrapy/base_spider.py
yolile: done

kingfisher_scrapy/settings.py
yolile: I agree with this and I already changed it to None, but now I'm not sure how to test it :persevere:

kingfisher_scrapy/base_spider.py
```diff
@@ -147,6 +152,12 @@ def from_crawler(cls, crawler, *args, **kwargs):
             except ValueError as e:
                 raise SpiderArgumentError(f'spider argument `until_date`: invalid date value: {e}')
 
+        if crawler.settings['DATABASE_URL']:
+            if not spider.crawl_time:
+                raise SpiderArgumentError('When using `DATABASE_URL` the spider argument `crawl_time` must be set')
+            if spider.compile_releases and hasattr(spider, 'data_type') and 'record' in spider.data_type:
+                raise SpiderArgumentError('The compile_releases flag can only be set if the spider returns releases.')
```
jpmckinney, line 159: Following same format as other errors in this method. Use `getattr` instead of `hasattr` + attribute access.

```suggestion
                raise SpiderArgumentError('spider argument `crawl_time`: can't be blank if `DATABASE_URL` is set')
            if spider.compile_releases and 'record' in getattr(spider, 'data_type', ''):
                raise SpiderArgumentError("spider argument `compile_releases`: can't be set if spider returns records")
```

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,153 @@ def _write_file(self, path, data):
...
+        if date_format == valid_formats['datetime']:
+            return date[:19]
+        if date_format == valid_formats['date']:
+            return date[:10]
+        if date_format == valid_formats['year-month']:
+            return date[:7]
+        return date[:4]
+
+    def spider_opened(self, spider):
+        self.crawl_directory = spider.crawl_time.strftime('%Y%m%d_%H%M%S')
```
jpmckinney, line 194: Can we move this to spider_closed and avoid a class attribute?

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,153 @@ def _write_file(self, path, data):
...
+                    spider.from_date = datetime.datetime.strptime(self.format_from_date(from_date, spider.date_format,
+                                                                                        spider.VALID_DATE_FORMATS),
+                                                                  spider.date_format)
+            self.connection.commit()
+        finally:
+            self.cursor.close()
+            self.connection.close()
+
+    def spider_closed(self, spider, reason):
+        if reason != 'finished':
```
jpmckinney, line 216: Should we allow 'sample' link in KingfisherProcessAPI?

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,153 @@ def _write_file(self, path, data):
...
+            return
+
+        if spider.compile:
```
jpmckinney, line 219: This should be `spider.compile_releases` (and again later).

kingfisher_scrapy/settings.py
```diff
@@ -122,6 +123,8 @@
 KINGFISHER_PLUCK_PATH = os.getenv('KINGFISHER_PLUCK_PATH', '')
 KINGFISHER_PLUCK_MAX_BYTES = None
 
+# To store items into a local Postgres data base
```
jpmckinney, line 126:
```suggestion
# To store items into a PostgreSQL database.
```

kingfisher_scrapy/settings.py
jpmckinney: Which test is it that needs to set `DATABASE_URL`? I can have a look.

kingfisher_scrapy/extensions.py
```diff
@@ -133,6 +139,153 @@ def _write_file(self, path, data):
...
+
+            # If there is not a from_date from the command line, get the most recent date in the spider's data table.
+            if not spider.from_date:
+                self.logger.info('Getting the date from which to resume the crawl (if any)')
+                self.execute("SELECT max(data->>'date') FROM {table}", table=spider.name)
+                from_date = self.cursor.fetchone()[0]
+                if from_date:
+                    spider.from_date = datetime.datetime.strptime(self.format_from_date(from_date, spider.date_format,
+                                                                                        spider.VALID_DATE_FORMATS),
+                                                                  spider.date_format)
```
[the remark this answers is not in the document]
yolile: This works, but we also need to convert `from_date` to datetime first:
````python
                if from_date:
                   # the date in the database is in OCDS dates format
                    from_date = datetime.strptime(from_date, '%Y-%m-%dT%H:%M:%S%z')
                    spider.from_date = datetime.strptime(datetime.strftime(from_date, spider.date_format),
                                                         spider.date_format)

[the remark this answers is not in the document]
jpmckinney: When the date is cast to `timestamptz`, does the PostgreSQL adapter not make it a Python datetime already?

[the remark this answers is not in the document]
yolile: true, I missed that part in the query!

kingfisher_scrapy/settings.py
yolile: We need a `DATABASE_URL` to test `spider_opened` and `spider_closed`. When this was a command we used an env variable, we can still do the same, but for testing it locally we will need to set the same env variable to make the tests work locally.

jpmckinney: Can we not use `get_crawler` from `scrapy.utils.test`, which accepts a `settings` argument, like we do in `tests/__init__.py`?

yolile: aha, but with which value?

jpmckinney: Oh, right, we'll need to read an environment variable.

yolile: aha, and the environment variable name will be set only in the tests? And if the required environment variable doesn't exist the test will fail and that is how the developer will know about that environment variable?

jpmckinney: We don't often document how to run tests. I typically refer to `ci.yml` to see what the test requirements are. We can either do that, or add a page to the docs on how to run tests.

jpmckinney: Or, we can restore `settings.py` and hope we never accidentally set the environment variable in production.

yolile: > We don't often document how to run tests. I typically refer to ci.yml to see what the test requirements are

I'm happy with this, as actually a lot of tests will fail if the environmental variable is always set in settings (eg because we test that if `DATABASE_URL` is set, `crawl_time` must be set too in the base class)

jpmckinney (dismissed):
I made one commit to reduce a bit of duplication.

jpmckinney:
The tests don't cover the case where the database returns a `from_date`? https://coveralls.io/builds/40061675/source?filename=kingfisher_scrapy%2Fextensions.py

yolile:
@jpmckinney added now

jpmckinney (approved):

PR #744, opened 2021-06-21
Add kaduna state spiders

### LINKED ISSUE #742: Kaduna State Budeshi Spider
https://www.budeshi.ng/kadppa/api

**Link to CRM issue**
CRM-7270

**Priority**
this week

cecicasco:
Closes #742


yolile (changes requested):
It seems that we can have a base class with the `start_request` method and the `base_url` class variable. method. We should also rename the spiders with `nigeria_` as prefix

cecicasco:
> It seems that we can have a base class with the `start_request` method and the `base_url` class variable. method. We should also rename the spiders with `nigeria_` as prefix

perfect..

docs/spiders.rst
```diff
@@ -671,6 +671,23 @@ Italy
 
    scrapy crawl italy
 
+Kaduna
```
yolile, line 674: @cecicasco we nees to update this too

yolile (approved):

PR #756, opened 2021-07-07
nigeria_releases: check for releases list with null values

### LINKED ISSUE #738: nigeria_budeshi_releases: null releases
There are some release packages eg https://budeshi.ng/api/releases/15394/tender that have "null" in the releases list. We should filter them, as now it makes kingfisher process fail
image

cc @odscrachel

yolile:
closes #738

### DIFF
--- a/kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
+++ b/kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
@@ -24,7 +24,7 @@ class NigeriaBudeshiReleases(NigeriaBudeshiBase):
 
     def parse(self, response):
         data = response.json()
-        # some responses includes a release list with null objects, eg:
+        # some responses include a release list with null objects, eg:
         #   "releases": [
         #     null
         #   ]
--- a/kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
+++ b/kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
@@ -28,7 +28,7 @@ class NigeriaBudeshiReleases(NigeriaBudeshiBase):
         #   "releases": [
         #     null
         #   ]
-        if 'releases' in data and data['releases'] == [None]:
+        if data['releases'] == [None]:
             yield self.build_file_error_from_response(response, errors=data)
         else:
             yield from super().parse(response)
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
```diff
@@ -21,3 +21,14 @@ def build_urls(self, project):
...
+    def parse(self, response):
+        data = response.json()
+        # some responses include a release list with null objects, eg:
+        #   "releases": [
+        #     null
+        #   ]
+        if 'releases' in data and data['releases'] and any(release is None for release in data['releases']):
+            yield self.build_file_error_from_response(response, errors=data)
+        else:
+            yield from super().parse(response)
```
jpmckinney, line 34: Do we want to reject the data entirely, or do:

```suggestion
        if 'releases' in data:
            data['releases'] = filter(None, data['releases'])
            yield self.build_file_from_response(response, data_type=self.data_type, data=data)
```

jpmckinney: Also, do we need the `if`? If all responses are expected to set it, then we do not need it.

yolile: Actually, all the cases that I've seen are with just one null object, and if we do the `filter` we will end up with empty packages, don't we?

jpmckinney: Let's simplify the `if` then - the current code suggests that some entries in the list might not be null.

yolile: > Let's simplify the if then to just if not data['releases']

That will not work as the list is `[None]`, I will change it to `data['releases'][0] is None`

jpmckinney: Yes - I realized before going to asleep, lol. Even clearer is `data['releases'] == [None]`

kingfisher_scrapy/spiders/nigeria_budeshi_releases.py
```diff
@@ -21,3 +21,14 @@ def build_urls(self, project):
...
         for tag in ('planning', 'tender', 'award', 'contract'):
             yield self.build_request(url.format(id=project['id'], tag=tag), formatter=components(-2))
+
+    def parse(self, response):
+        data = response.json()
+        # some responses include a release list with null objects, eg:
+        #   "releases": [
+        #     null
+        #   ]
+        if 'releases' in data and data['releases'] == [None]:
```
jpmckinney, line 31: I'm still curious why we have `'releases' in data`. (1) Has it been observed to be missing? (2) If so, do we really want to process the file normally if missing?

I think better to remove it, if possible.

yolile: true, no, I haven't seen that, removed

jpmckinney (approved):
Great :) In general, my experience writing scrapers is that the code should be very simple and strict. It's often the opposite to what you would do when writing a library (where you want to be more accepting of different inputs). This is because crawled data is very unpredictable, so it's better to fail fast with strict code, so that unexpected cases don't pass through.

PR #774, opened 2021-08-04
add nigeria cross river state spider

### LINKED ISSUE #766: Cross River State (Nigeria)
http://ocds.dppib-crsgov.org/calabar/dashboard/records

http://18.202.236.131:8101/records/json-download/recordPackage_2020_2_1

The bulk download URLs seem to follow a consistent pattern:

http://18.202.236.131:8101/records/json-download/recordPackage_[year]_[month]_1

**Link to CRM issue**
CRM-7426

**Priority**
This week

**Data structure**
Structurally valid, with some additional fields: https://standard.open-contracting.org/review/data/9cec7d9f-dd5c-4ca7-a63f-cc33f8e7ab2a

Ravf95:
Closes #766

### DIFF
--- a/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
+++ b/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
@@ -1,0 +1,21 @@
+from abc import abstractmethod
+
+from kingfisher_scrapy.base_spider import PeriodicSpider
+from kingfisher_scrapy.util import components, join, parameters
+
+
+class NigeriaCrossRiverBase(PeriodicSpider):
+    # SimpleSpider
+    base_url = 'http://ocdsapi.dppib-crsgov.org/api/ocdsAPI/'
+
+    # BaseSpider
+    date_format = 'year-month'
+    default_from_date = '2020-02'
+    default_until_date = '2020-11'
+
+    def get_formatter(self):
+        return join(components(-1), parameters('year', 'month'))
+
+    @abstractmethod
+    def build_urls(self, date):
+        pass
--- a/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
+++ b/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
@@ -11,7 +11,6 @@ class NigeriaCrossRiverBase(PeriodicSpider):
     # BaseSpider
     date_format = 'year-month'
     default_from_date = '2020-02'
-    default_until_date = '2020-11'
 
     def get_formatter(self):
         return join(components(-1), parameters('year', 'month'))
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/spiders/nigeria_cross_river_records.py
```diff
@@ -0,0 +1,18 @@
+from kingfisher_scrapy.spiders.nigeria_cross_river_base import NigeriaCrossRiverBase
+
+
+class NigeriaCrossRiverRecords(NigeriaCrossRiverBase):
+    """
+    Domain
+      Cross River Nigeria
+    API documentation
```
yolile, line 8: same as the other we should add the Spider arguments for `from_date` and `until_date`

kingfisher_scrapy/spiders/nigeria_cross_river_base.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class NigeriaCrossRiverBase(PeriodicSpider):
+    # SimpleSpider
+    base_url = 'http://ocdsapi.dppib-crsgov.org/api/ocdsAPI/'
+
+    # BaseSpider
+    date_format = 'year-month'
+    default_from_date = '2020-02'
+    default_until_date = '2020-11'
```
yolile, line 14: how do you know these are the default dates?

yolile (changes requested):

kingfisher_scrapy/spiders/nigeria_cross_river_records.py
Ravf95: I fix it in the next commit

kingfisher_scrapy/spiders/nigeria_cross_river_base.py
Ravf95: I take `from_date` and `until_date` from the monthly download list of
http://ocds.dppib-crsgov.org/ReleasePackageViewModel_ListView/

yolile: Then, instead of implementing this as a `PeriodicSpider`, I think that we should get the list of the available dates from that page. Right now the spider will try to get the data for all the months from February to November, but I see in that list that there are some unavailable months, as May, July, September, etc. So we should a) Document ttp://ocds.dppib-crsgov.org/ReleasePackageViewModel_ListView/ as "Bulk download documentation" and b) Reimplement the spider to get the available months and years from there.

yolile (changes requested):

Ravf95:
Yes! We can try to obtain the available dates of the page, it would take more time because it is a page programmed with ASP.NET and it is more complex to scrape it, I did not do it like that due to the priority of the issue.
We should also think about pagination, currently, we only have one page of results but in the future, we could have more, should I implement this functionality now anyway? I won't be able to test it

yolile:
Hmm, ok, so let's merge this now (you will need to resolve the conflict files) but please open a new issue to fix it later.

yolile (approved):

Issue #775, opened 2021-08-06
Update Nigeria Cross River Spider


Ravf95:
In [this](https://github.com/open-contracting/kingfisher-collect/pull/774) pull we discussed reimplement the spider to get the available months and years from [here](http://ocds.dppib-crsgov.org/ReleasePackageViewModel_ListView/) for request only the available dates. This feature requires a scraper from ASP.NET programmed page.

**Link to github issue**
#766

**Priority**
anytime

yolile:
blocked due to https://github.com/open-contracting/kingfisher-collect/pull/786#issuecomment-910888903

Ravf95:
The publisher made changes, now we have API endpoint http://ocdsapi.dppib-crsgov.org/api/ocdsAPI/getAvailableReleasesSummary for request available dates

PR #786, opened 2021-08-31
update nigeria cross river spider

Ravf95:
I’m working to try to get the dates and see how we can simulated click on future pages using the form from the nigeria cross river publication data site, currently, [the website](http://ocds.dppib-crsgov.org/ReleasePackageViewModel_ListView/) is up but has a data load problem, [the API request method](https://github.com/open-contracting/kingfisher-collect/issues/766#issuecomment-890683395) doesn't work.

The site doesn't use these inputs for the pagination:
```
__EVENTTARGET
__EVENTARGUMENT
```

The pagination work with the javascript onclick function: `ASPx.GVPagerOnClick('...','PN<page-number>')`
I think we can scrape this with the[ selenium module](https://selenium-python.readthedocs.io/waits.html) for future pages.

With this update, we can scrape the dates and request them. Ref #775

### DIFF
--- a/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
+++ b/kingfisher_scrapy/spiders/nigeria_cross_river_base.py
@@ -1,16 +1,38 @@
 from abc import abstractmethod
+from datetime import datetime
 
-from kingfisher_scrapy.base_spider import PeriodicSpider
-from kingfisher_scrapy.util import components, join, parameters
+import scrapy
+
+from kingfisher_scrapy.base_spider import SimpleSpider
+from kingfisher_scrapy.util import components, handle_http_error, join, parameters
 
 
-class NigeriaCrossRiverBase(PeriodicSpider):
+class NigeriaCrossRiverBase(SimpleSpider):
     # SimpleSpider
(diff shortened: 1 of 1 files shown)


yolile:
I don't see any packages or records on their site anymore http://ocds.dppib-crsgov.org/ReleasePackageViewModel_ListView/ :disappointed:

yolile:
> I think we can scrape this with the selenium module for future pages.

I'm not sure if we can use this as we will need a browser, so if there is no other way, we will need to ask the publisher to have a service for listing the available files/dates.

jpmckinney:
Yes, although we can do it, our broader goal is for data to be retrievable without a significant effort. So, better to inform the publisher and wait for a better publication.

yolile:
@Ravf95 could you let the corresponding helpdesk analyst (in CRM-7426) know that this needs to be informed to the publisher? I'm closing this PR in favor of that. Thanks

Ravf95:
Yes, I updated the issue

Ravf95:
Currently, I can see again the data on the publisher website but the [API](https://github.com/open-contracting/kingfisher-collect/issues/766#issuecomment-890683395) doesn't work. I update [crm issue](https://crm.open-contracting.org/issues/7426) and also I ask about other methods to get all available dates

PR #791, opened 2021-09-01
update kyrgyz spider

### LINKED ISSUE #785: Update Kyrgyz Spider
In the ECA Checkin on 2021-08-31 we noted that the Kyrgyz spider should be updated with the following URL: http://ocds.zakupki.gov.kg/api-export

I've just tested it and I am getting a 404 error, so I'm chasing this on the CRM as well.

Ravf95:
Closes #785

### DIFF
--- a/kingfisher_scrapy/spiders/kyrgyzstan.py
+++ b/kingfisher_scrapy/spiders/kyrgyzstan.py
@@ -12,6 +12,7 @@ class Kyrgyzstan(LinksSpider):
       The planning endpoint is not implemented because is not completely in OCDS.
     Bulk download documentation
       http://ocds.zakupki.gov.kg/dashboard/weekly and then to 'API & EXPORT'.
+      Direct access to http://ocds.zakupki.gov.kg/api-export doesn't work.
     Swagger API documentation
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/kyrgyzstan.py
```diff
@@ -11,8 +11,7 @@ class Kyrgyzstan(LinksSpider):
     Caveats
       The planning endpoint is not implemented because is not completely in OCDS.
     Bulk download documentation
-      https://analytics.zakupki.gov.kg/dashboard/weekly and then to 'API & EXPORT'.
-      Direct access to https://analytics.zakupki.gov.kg/api-export doesn't work.
```
yolile, line 15: @Ravf95 this is still true, we just need to change the URLs

yolile (changes requested):

kingfisher_scrapy/spiders/kyrgyzstan.py
Ravf95: The change has available in the latest commit!

yolile (approved):

PR #792, opened 2021-09-03
update kenya makueni spider

### LINKED ISSUE #788: KenyaMakueni: Can't use IndexSpider
The count endpoint returns the number of results *on that page* not in the dataset.

For example, https://opencontracting.makueni.go.ke/api/ocds/release/count?pageSize=1 returns 1 🙃 

The default `pageSize` is 300, so we were always getting 300 from https://opencontracting.makueni.go.ke/api/ocds/release/count

Right now, there are fewer than 1000, so we can get all the data by increasing the pageSize to 1000 (the maximum), but once they exceed 1000 releases we will need a new strategy, so we should implement it now.

Ravf95:
Closes #788

### DIFF
--- a/kingfisher_scrapy/spiders/kenya_makueni.py
+++ b/kingfisher_scrapy/spiders/kenya_makueni.py
@@ -1,9 +1,8 @@
-import scrapy
+from kingfisher_scrapy.base_spider import SimpleSpider
+from kingfisher_scrapy.util import handle_http_error, parameters
 
-from kingfisher_scrapy.base_spider import IndexSpider
 
-
-class KenyaMakueni(IndexSpider):
+class KenyaMakueni(SimpleSpider):
     """
     Domain
       Makueni County
@@ -18,14 +17,21 @@ class KenyaMakueni(IndexSpider):
     # SimpleSpider
     data_type = 'release_package'
 
-    # IndexSpider
-    count_pointer = ''
-    limit = 1000  # > 1000 causes "must be between 1 and 1000"
-    use_page = True
-    start_page = 0
-    param_page = 'pageNumber'
-    base_url = f'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={limit}'
+    step = 100
+    url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={0}&pageNumber={1}'
 
     def start_requests(self):
-        url = f'https://opencontracting.makueni.go.ke/api/ocds/release/count?pageSize=1000'
-        yield scrapy.Request(url, meta={'file_name': 'count.json'}, callback=self.parse_list)
+        yield self.build_request(self.url.format(self.step, 0),
+                                 self.get_formatter(), meta={'page': 0}, callback=self.parse_list)
+
+    @handle_http_error
+    def parse_list(self, response):
+        page = response.request.meta['page']
+        yield self.build_file_from_response(response, data_type=self.data_type)
+
+        if len(response.json()) > 0:
+            yield self.build_request(self.url.format(self.step, page + 1),
+                                     self.get_formatter(), meta={'page': page + 1}, callback=self.parse_list)
+
+    def get_formatter(self):
+        return parameters('pageNumber')
--- a/kingfisher_scrapy/spiders/kenya_makueni.py
+++ b/kingfisher_scrapy/spiders/kenya_makueni.py
@@ -18,18 +18,16 @@ class KenyaMakueni(SimpleSpider):
     data_type = 'release_package'
 
     def start_requests(self):
-        yield from self.request_page()
+        yield from self.request_page(0)
 
     @handle_http_error
     def parse(self, response):
         yield from super().parse(response)
 
-        if len(response.json()) > 0:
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -18,14 +17,21 @@ class KenyaMakueni(IndexSpider):
...
+    def parse_list(self, response):
+        page = response.request.meta['page']
+        yield self.build_file_from_response(response, data_type=self.data_type)
+
+        if len(response.json()) > 0:
+            yield self.build_request(self.url.format(self.step, page + 1),
+                                     self.get_formatter(), meta={'page': page + 1}, callback=self.parse_list)
+
+    def get_formatter(self):
+        return parameters('pageNumber')
```
jpmckinney, line 37: There's some repetition across the methods. I suggest extracting the build request call to a separate method, that accepts a page. If we do that, we can also in-line the url, step and formatter, so that they aren't class attributes or separate methods.

Also, we can rename parse_list to parse, so that we just call `super()`. Scrapy's default callback is `parse`, so that let's us also remove the explicit callback keyword argument.

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -18,14 +17,21 @@ class KenyaMakueni(IndexSpider):
...
     data_type = 'release_package'
 
-    # IndexSpider
-    count_pointer = ''
-    limit = 1000  # > 1000 causes "must be between 1 and 1000"
-    use_page = True
-    start_page = 0
-    param_page = 'pageNumber'
-    base_url = f'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={limit}'
+    step = 100
```
jpmckinney, line 20: Increase to 1000 once you've tested that the pagination works.

kingfisher_scrapy/spiders/kenya_makueni.py
Ravf95: Thanks for the suggestions, they're were applied

kingfisher_scrapy/spiders/kenya_makueni.py
Ravf95: Yes, I change the value

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -18,14 +17,19 @@ class KenyaMakueni(IndexSpider):
...
+
+        if len(response.json()) > 0:
+            page = response.request.meta['page']
+            yield from self.request_page(page + 1)
+
+    def request_page(self, page=0):
+        step = 1000
+        url = 'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize={0}&pageNumber={1}'
+
+        yield self.build_request(url.format(step, page), parameters('pageNumber'), meta={'page': page})
```
jpmckinney, line 35: Construct the URL in fewer steps:

```suggestion
        url = f'https://opencontracting.makueni.go.ke/api/ocds/package/all?pageSize=1000&pageNumber={page}'
        yield self.build_request(url, parameters('pageNumber'), meta={'page': page})
```

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -18,14 +17,19 @@ class KenyaMakueni(IndexSpider):
...
     def start_requests(self):
-        url = f'https://opencontracting.makueni.go.ke/api/ocds/release/count?pageSize=1000'
-        yield scrapy.Request(url, meta={'file_name': 'count.json'}, callback=self.parse_list)
+        yield from self.request_page()
+
+    @handle_http_error
+    def parse(self, response):
+        yield from super().parse(response)
+
+        if len(response.json()) > 0:
```
jpmckinney, line 27: `[]` is already falsy.

```suggestion
        if response.json():
```

kingfisher_scrapy/spiders/kenya_makueni.py
```diff
@@ -18,14 +17,19 @@ class KenyaMakueni(IndexSpider):
...
+            page = response.request.meta['page']
+            yield from self.request_page(page + 1)
+
+    def request_page(self, page=0):
```
jpmckinney, line 31: It might be better to remove the default value, and use `self.request_page(0)` in `start_requests`. That way, it's impossible to accidentally call the method without the desired page.

kingfisher_scrapy/spiders/kenya_makueni.py
Ravf95: Done

kingfisher_scrapy/spiders/kenya_makueni.py
Ravf95: Yes, you're right

kingfisher_scrapy/spiders/kenya_makueni.py
Ravf95: Yes, thanks for the suggestions, the changes are available in 5682767

jpmckinney (approved):

PR #800, opened 2021-09-20
Fix Canada Buy and Sell docs

kindly:
Class name needs updating in docs


jpmckinney:
Done in 71d3430 with a bunch of others (using `scrapy updatedocs` command).

Issue #414, opened 2020-06-04
Update bulk download scrapers to scrape from pages instead of using hard coded URLs


yolile:
This includes at least : 

- [x] Chile https://desarrolladores.mercadopublico.cl/OCDS/DescargaMasiva
- [x] Canada Buyandsell https://buyandsell.gc.ca/procurement-data/open-contracting-data-standard-pilot/download-ocds-pilot-data
- [x] Honduras http://www.contratacionesabiertas.gob.hn/descargas/
- [x] Uruguay https://www.gub.uy/agencia-compras-contrataciones-estado/datos-y-estadisticas/datos/open-contracting
- [x] Nepal http://ppip.gov.np/downloads

nativaldezt:
So, the idea is to get the download URLs from the page instead of assuming it is a certain way?
This is to prevent the spider from stopping working if the link changes? that is more likely to happen than the page being changed?

yolile:
> So, the idea is to get the download URLs from the page instead of assuming it is a certain way?

Yes

> This is to prevent the spider from stopping working if the link changes?

Yes, but also to get a better chance to know if they have added more links and download the new links too, not just the fixed ones that we may be aware of

yolile:
update, for Honduras I see that they actually have an endpoint that lists all the available URLs, so we can use that  one http://www.contratacionesabiertas.gob.hn/api/v1/descargas/

Ravf95:
Hi @yolile.  I started with the chile_compras_bulk spider and I see that the load dates in https://desarrolladores.mercadopublico.cl/OCDS/DescargaMasiva  work with a request to https://api.mercadopublico.cl/APISOCDS/Utilidades/DescargaMasiva and this calls to the javascript function [ListItems(initYear)](https://api.mercadopublico.cl/APISOCDS/Utilidades/Scripts/descarga-masiva.js) to load data in `div #contentList`, so the load data requires an initial year for iterate to current year. I think that the spider doesn't require changes, what do you think?

yolile:
Yes, makes sense, and they are also documeting the use as "En caso de querer descargar los archivos mediante programas, es posible hacerlo mediante URLs en las que se modifique la fecha del archivo de la forma: https://ocds.blob.core.windows.net/ocds/yyyymm.zip", so we are already using the service as expected, as they consintently publish a file per year-month.

yolile:
I think that we can also skip Canada Buyandsell, as they have more links than the ones listed on their page.

[...]

PR #804, opened 2021-09-24
update honduras portal bulk spider

Ravf95:
Ref #414

### DIFF
--- a/kingfisher_scrapy/spiders/honduras_portal_bulk.py
+++ b/kingfisher_scrapy/spiders/honduras_portal_bulk.py
@@ -58,20 +58,41 @@ class HondurasPortalBulk(SimpleSpider):
     @classmethod
     def from_crawler(cls, crawler, publisher=None, system=None, *args, **kwargs):
         spider = super().from_crawler(crawler, publisher=publisher, system=system, *args, **kwargs)
-        if publisher and spider.publisher not in spider.available_publishers.keys():
+        if publisher and spider.publisher not in spider.available_publishers:
             raise SpiderArgumentError(f'spider argument `publisher`: {spider.publisher!r} not recognized')
 
         if system:
             if spider.publisher != 'oncae':
                 raise SpiderArgumentError(f'spider argument `system` is not supported for publisher: '
                                           f'{spider.publisher!r}')
-            if spider.system not in spider.available_systems.keys():
+            if spider.system not in spider.available_systems:
                 raise SpiderArgumentError(f'spider argument `system`: {spider.system!r} not recognized')
 
         return spider
 
     @handle_http_error
     def parse_list(self, response):
+        # The system names are:
+        #    - oficina_normativa_honducompras-1
+        #    - oficina_normativa_catalogo-electronico
+        #    - oficina_normativa_difusion-directa-contrato
+        #    - secretaria_de_fin_HN.SIAFI2
+        #
+        # An example of expected data is:
+        # [
+        #   {
+        #    "urls": {
+        #      "csv": "https://www.contratacionesabiertas.gob.hn/api/v1/descargas/<system_name>_<year>_<month>.zip",
+        #      "md5": "https://www.contratacionesabiertas.gob.hn/api/v1/descargas/<system_name>_<year>_<month>.md5",
+        #      "json": "https://www.contratacionesabiertas.gob.hn/api/v1/descargas/<system_name>_<year>_<month>.json",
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/honduras_portal_bulk.py
```diff
@@ -41,39 +45,48 @@ class HondurasPortalBulk(PeriodicSpider):
...
-    available_systems = {'HC1': 'honducompras-1', 'CE': 'catalogo-electronico', 'DDC': 'difusion-directa-contrato'}
+    def start_requests(self):
+        url = 'http://www.contratacionesabiertas.gob.hn/api/v1/descargas/?format=json'
+        yield scrapy.Request(url, meta={'file_name': 'list.json'}, callback=self.parse_list)
 
     @classmethod
     def from_crawler(cls, crawler, publisher=None, system=None, *args, **kwargs):
         spider = super().from_crawler(crawler, publisher=publisher, system=system, *args, **kwargs)
-        if publisher and spider.publisher not in spider.available_publishers:
+        if publisher and spider.publisher not in spider.available_publishers.keys():
```
yolile, line 61:
```suggestion
        if publisher and spider.publisher not in spider.available_publishers:
```
.keys() is the default behavior

kingfisher_scrapy/spiders/honduras_portal_bulk.py
```diff
@@ -41,39 +45,48 @@ class HondurasPortalBulk(PeriodicSpider):
...
             raise SpiderArgumentError(f'spider argument `publisher`: {spider.publisher!r} not recognized')
 
         if system:
             if spider.publisher != 'oncae':
                 raise SpiderArgumentError(f'spider argument `system` is not supported for publisher: '
                                           f'{spider.publisher!r}')
-            if spider.system not in spider.available_systems:
+            if spider.system not in spider.available_systems.keys():
```
yolile, line 68:
```suggestion
            if spider.system not in spider.available_systems:
```

.keys() is the default behavior

kingfisher_scrapy/spiders/honduras_portal_bulk.py
```diff
@@ -41,39 +45,48 @@ class HondurasPortalBulk(PeriodicSpider):
...
-                                              f"{self.available_systems[system]}_{date.year}_{date.month:02d}.json")
-            else:
-                yield self.pattern.format(f"{self.available_publishers[publisher]}_"
-                                          f"{date.year}_{date.month:02d}.json")
+            if publisher == self.available_publishers['oncae']:
+                system = item['sistema']
+                if self.system and system != self.available_systems.get(self.system):
+                    continue
+
+            date = datetime(int(item['year']), int(item['month']), 1)
```
yolile, line 87: You can move this inside the if block, as the date is only needed if the from_date and until_date are set

kingfisher_scrapy/spiders/honduras_portal_bulk.py
```diff
@@ -41,39 +45,48 @@ class HondurasPortalBulk(PeriodicSpider):
...
         return spider
 
-    def build_urls(self, date):
-        for publisher in self.available_publishers:
-            if self.publisher and publisher != self.publisher:
+    @handle_http_error
+    def parse_list(self, response):
+        formatter = components(-1)
+
+        for item in response.json():
```
yolile, line 77: Could you add a comment on how each item looks so it is easier for another developer to understand the code and filters? eg 
``` python
# "Each item looks like:  {
#    "urls": {
#      "csv": "...",
#      "md5": "...",
#      "json": "...",
#      "xlsx": "..."
#    },
#    "year": "2012",
#    "month": "01",
#    "sistema": "Sistema de Informacion Financiero Integrado v2.0",
#    "publicador": "Secretaria de Finanzas de Honduras"
#  }

yolile (changes requested):

kingfisher_scrapy/spiders/honduras_portal_bulk.py
Ravf95: Yes, you're right!. I will remove it on the next commit

kingfisher_scrapy/spiders/honduras_portal_bulk.py
Ravf95: Yes, you're right!. I will remove it on the next commit

kingfisher_scrapy/spiders/honduras_portal_bulk.py
Ravf95: Yes, I will add a comment

yolile (approved):

PR #810, opened 2021-09-30
Change response manage to get codes

### LINKED ISSUE #673: Indonesia - opentender.net

cecicasco:
closes #673


yolile (changes requested):
I think that we need to use "format=json" as an additional query param now, eg: https://opentender.net/api/master/lpse?year=2020&format=json

kingfisher_scrapy/spiders/indonesia_opentender.py
```diff
@@ -27,8 +27,7 @@ class IndonesiaOpentender(CompressedFileSpider, PeriodicSpider):
     data_type = 'release_package'
 
     # PeriodicSpider
-    pattern = base_url + 'master/lpse?year={}'
-    formatter = staticmethod(components(-1))
```
yolile, line 31: The `formatter` variable is needed for `PeriodicSpider`

yolile (approved):

PR #811, opened 2021-09-30
add nigeria gombe state spider

Ravf95:
Closes #805

PR #826, opened 2021-10-06
add base class for nigeria portal

### LINKED ISSUE #815: Base class for Nigeria Portal and Gombe State
We can write a base class for Nigeria Portal and Gombe State.
The base class can reuse the current implementation of `NigeriaPortal class`.

Ref  https://github.com/open-contracting/kingfisher-collect/issues/811#issuecomment-931833637

**Priority**
anytime

Ravf95:
Closes #815


Ravf95:
Hi @yolile. Yes, this still working with the changes. I don't take the previous implementation (see pull #811) for both spiders because is better (less request send) to download one file that contains all releases (array of releases) instead of download the release file one by one. Probably we can reuse the previous implementation to add support to download by date. What do you think?

PR #849, opened 2021-10-29
update india_civic_data_lab spider

### LINKED ISSUE #821: CivicDataLab: Use GitHub API instead of parsing HTML
This is low priority.

Ravf95:
Closes #821

### DIFF
--- a/kingfisher_scrapy/spiders/india_civic_data_lab.py
+++ b/kingfisher_scrapy/spiders/india_civic_data_lab.py
@@ -25,12 +25,18 @@ class IndiaCivicDataLab(SimpleSpider):
     # SimpleSpider
     data_type = 'release_package'
 
+    # GitHub repository
+    github_repo = 'CivicDataLab/himachal-pradesh-health-procurement-OCDS'
+
     def start_requests(self):
-        url = 'https://api.github.com/repos/CivicDataLab/himachal-pradesh-health-procurement-OCDS/git/trees/master'
-        yield scrapy.Request(url, meta={'file_name': 'response.json'}, callback=self.parse_file_list)
+        url = f'https://api.github.com/repos/{self.github_repo}/git/trees/master'
+        yield scrapy.Request(url, meta={'file_name': 'response.json'}, callback=self.parse_list)
 
     @handle_http_error
-    def parse_file_list(self, response):
+    def parse_list(self, response):
+        # The parser uses the GitHub API to list the files from the repository and
+        # then downloads them with a non-API url to avoid quota/rate limitations
+
         data = response.json()
         # The data looks like:
         # {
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/india_civic_data_lab.py
```diff
@@ -24,14 +26,37 @@ class IndiaCivicDataLab(SimpleSpider):
...
+        #   "sha": ...
+        #   "url": ...
+        #   "tree": [ ... ],
+        #   "truncated": ...
+        # }
+
+        for node in data['tree']:
+            file_name = node['path']
+            if file_name.endswith('.xlsx'):
+                yield scrapy.Request(node['url'], meta={'file_name': file_name}, callback=self.parse_file_content)
```
yolile, line 46: I'm not sure we need to use the GitHub API to download the file as well. I think it is ok to use the API to get the file name, and then use the normal download URL to download the file, eg, with just one extra method (avoiding `parse_file_content`): 
```` python
    @handle_http_error
        data = response.json()
        for node in data['tree']:
            file_name = node['path']
            if file_name.endswith('.xlsx'):
                url = f'https://github.com/{self.github_repo}/raw/master/{file_name}?raw=true'
                yield scrapy.Request(url, meta={'file_name': file_name})
```` 
where ``github_repo = 'CivicDataLab/himachal-pradesh-health-procurement-OCDS'`` 

Or @jpmckinney what do you think? Did you mean to use the API for both, getting the files list and downloading them?

Comment (jpmckinney) [reply kingfisher_scrapy/spiders/india_civic_data_lab.py:46]: Just getting the file list is enough. Using the API to download will consume more of the GitHub quota / rate limit, so it's safer to download from the non-API URLs.

Comment (Ravf95) [reply kingfisher_scrapy/spiders/india_civic_data_lab.py:46]: Thanks for the suggestions. I will push the changes in the next commit!

**Review by yolile — APPROVED** [review]

File: kingfisher_scrapy/spiders/india_civic_data_lab.py [hunk 2]
```diff
@@ -23,15 +23,29 @@ class IndiaCivicDataLab(SimpleSpider):
...
+        # then downloads them with a non-API url to avoid quota/rate limitations
+
+        data = response.json()
+        # The data looks like:
+        # {
+        #   "sha": ...
+        #   "url": ...
+        #   "tree": [ ... ],
+        #   "truncated": ...
+        # }
```
jpmckinney, line 45: We should instead link to the GitHub API documentation: https://docs.github.com/en/rest/reference/git#get-a-tree

jpmckinney: Also, we should check if `truncated` is true and yield an error if so, because it means we don't have the full tree.

jpmckinney (commented):
cc @yolile @Ravf95

kingfisher_scrapy/spiders/india_civic_data_lab.py
Ravf95: @jpmckinney @yolile About the truncated check. I don't sure how we do it. I think with `self.logger.error()` before the iteraration and then raise a exeption (maybe a new exepction class in exceptions.py), or do you suggest another error handling method?

jpmckinney: We can either log an error, yield a FileError, or raise an Exception. I guess raising an exception would be best, as otherwise a user might not notice the log messages.

Ravf95: Thanks, the changes are available in PR #856

PR #854, opened 2021-11-02
docs: fixs requirements to generate spiders, pipelines and extensions docs

yolile:
Otherwise, the readthedocs build fails eg https://readthedocs.org/projects/kingfisher-collect/builds/15163286/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,8 @@
+flattentool==0.14.0
+ijson==3.1.1
+jsonpointer==2.0
+pika==1.2.0
+psycopg2==2.8.6
+rarfile==3.1
+scrapy==2.5.1
 sphinx>2,<4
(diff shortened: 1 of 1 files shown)


docs/requirements.txt
```diff
@@ -1 +1,8 @@
-sphinx>2,<4
\ No newline at end of file
+flattentool==0.14.0
+ijson==3.1.1
+jsonpointer==2.0
+pika==1.2.0
+psycopg2==2.8.6
+rarfile==3.1
+scrapy==2.5.1
```
jpmckinney, line 7: Can we do `-r ../requirements.txt` ?

yolile: It seems so, updated now

jpmckinney (approved):
Interesting - I guess paths are relative to the directory from which `pip` is run!

PR #856, opened 2021-11-04
update india_civic_data_lab spider 

### LINKED ISSUE #821: CivicDataLab: Use GitHub API instead of parsing HTML

Ravf95:
Closes #821

### DIFF
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -232,4 +232,4 @@ class RetryDataErrorMiddleware:
                                 {'request': response.request, 'failures': retries, 'exception': exception})
             yield request
         else:
-           raise exception
+            raise exception
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -231,3 +231,5 @@ class RetryDataErrorMiddleware:
             spider.logger.debug('Retrying %(request)s (failed %(failures)d times): %(exception)s',
                                 {'request': response.request, 'failures': retries, 'exception': exception})
             yield request
+        else:
+           raise exception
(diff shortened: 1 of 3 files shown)


kingfisher_scrapy/spiders/india_civic_data_lab.py
```diff
@@ -36,13 +36,14 @@ def parse_list(self, response):
...
-        #   "truncated": ...
-        # }
+        # GitHub Tree API documentation
+        # https://docs.github.com/en/rest/reference/git#get-a-tree
+
+        # If truncated is true in the response then the repository has exceeded
+        # the maximum number of items can GitHub Tree API do handle
+        if data['truncated']:
+            self.logger.error('Truncated results in GitHub Tree API request.')
+            raise Exception()
```
yolile, line 46:
```suggestion
            raise KingfisherScrapyError('Truncated results returned when querying the file list from GitHub')
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -231,3 +231,5 @@ def process_spider_exception(self, response, exception, spider):
             spider.logger.debug('Retrying %(request)s (failed %(failures)d times): %(exception)s',
                                 {'request': response.request, 'failures': retries, 'exception': exception})
             yield request
+        else:
+           raise exception
```
yolile, line 235: FYI @jpmckinney I added this to the `RetryDataErrorMiddleware` otherwise the spider's exceptions are not raised!

yolile (changes requested):

kingfisher_scrapy/spidermiddlewares.py
jpmckinney: Oh, right! Did you check any other `process_spider_exception` handlers for the same problem?

yolile: Yes, and this is the only `process_spider_exception` handler we have so far so this change should be enough

yolile:
@jpmckinney let me know if it looks good for you now

jpmckinney (approved):

PR #859, opened 2021-11-05
KingfisherProcessAPI error callback

### LINKED ISSUE #672: RequestGenerationFailed: [<twisted.python.failure.Failure builtins.AttributeError: 'FileBodyProducer' object has no attrib...
I don't understand this error. I think `_task` is something that exists in `twisted/web/client.py`. The error is caught in `scrapy/utils/signal.py`.

To start, we can upgrade sentry_sdk to the latest version, to see if we can get a more useful error report.

Sentry Issue: [KINGFISHER-COLLECT-B](https://sentry.io/organizations/open-contracting-partnership/issues/2103341406/?referrer=github_integration)

```
RequestGenerationFailed: [<twisted.python.failure.Failure builtins.AttributeError: 'FileBodyProducer' object has no attribute '_task'>]

Error caught on signal handler: <bound method KingfisherProcessAPI.item_scraped of <kingfisher_scrapy.extensions.KingfisherProcessAPI object at 0x7f7f629c8438>>
```

yolile:
closes #672

### DIFF
--- a/kingfisher_scrapy/extensions.py
+++ b/kingfisher_scrapy/extensions.py
@@ -436,8 +436,9 @@ class KingfisherProcessAPI:
             # A return value is provided to ease testing.
             return response
 
-        def log_for_exception(error):
-            spider.logger.warning('%s failed (%s) with exception: %d', method, infix, error.getTraceback())
+        def log_for_exception(exception):
+            spider.logger.warning('%s failed (%s) with exception: %d', method, infix, exception.getTraceback())
+            raise exception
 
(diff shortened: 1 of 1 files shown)


jpmckinney (approved):
I'd normally ask to add a test, but this code is scheduled for removal once we switch to the new Process.

yolile:
There is a test already! Which failed  first because the test expected an exception, that is why I logged it and then raise the exception too, as expected

jpmckinney:
Ah, great! You can merge.

PR #870, opened 2021-11-24
chore: Add new exception for logging errors in spider init and from_crawler methods

### LINKED ISSUE #868: Spiders' from_crawler exceptions are not being logged
None of the exceptions raised in Spider's `from_crawler` methods are being logged. For example when running `scrapy crawl colombia_ani_records -a sample=asds` the output is https://collect.kingfisher.open-contracting.org/logs/kingfisher/colombia_ani_records/38ca0a504d3a11ec90ee0c9d92c523cb.log, with no further logging after the exception is raised (in this case the expected exception is https://github.com/open-contracting/kingfisher-collect/blob/60da8f1ef6c6bd66fab186beebe27eaf67df65b0/kingfisher_scrapy/base_spider.py#L182). 

I tried locally by deactivating the RetryMiddleware https://github.com/open-contracting/kingfisher-collect/blob/60da8f1ef6c6bd66fab186beebe27eaf67df65b0/kingfisher_scrapy/spidermiddlewares.py#L214 just in case, and also downgrading Scrapy, without any luck.

yolile:
closes #868

### DIFF
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -1,12 +1,21 @@
+from scrapy.utils import spider
+
+
 class KingfisherScrapyError(Exception):
     """Base class for exceptions from within this application"""
 
 
-class SpiderArgumentError(KingfisherScrapyError):
+class SpiderInitError(KingfisherScrapyError):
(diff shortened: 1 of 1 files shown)


yolile:
closed in favor of #871

PR #871, opened 2021-11-24
Add SpiderInitializationError base class

jpmckinney:
```
python -m scrapy crawl error -s SPIDER_MODULES=tests.spiders -a raise_from_crawler=True

...

2021-11-24 13:56:06 [error] INFO: Spider arguments: {'sample': None, 'note': None, 'from_date': None, 'until_date': None, 'crawl_time': None, 'keep_collection_open': None, 'package_pointer': None, 'release_pointer': None, 'truncate': None, 'compile_releases': None, 'raise_from_crawler': 'True', 'raise_init': False, 'raise_start_requests': False, 'raise_parse': False}
2021-11-24 13:56:06 [kingfisher_scrapy.exceptions] ERROR: message
  File "/Users/james/Sites/remote/open-contracting/kingfisher-collect/tests/spiders/error.py", line 35, in from_crawler
    raise SpiderArgumentError("message")
```

### DIFF
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -1,9 +1,11 @@
+import logging
+import traceback
+
+logger = logging.getLogger(__name__)
+
+
 class KingfisherScrapyError(Exception):
     """Base class for exceptions from within this application"""
-
-
-class SpiderArgumentError(KingfisherScrapyError):
-    """Raised when a spider argument's value is invalid"""
 
 
 class MissingEnvVarError(KingfisherScrapyError):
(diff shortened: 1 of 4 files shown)


kingfisher_scrapy/exceptions.py
```diff
@@ -22,5 +24,16 @@ class UnknownArchiveFormatError(KingfisherScrapyError):
     """Raised when the archive format of a file can't be determined from the filename"""
 
 
-class IncoherentConfigurationError(KingfisherScrapyError):
-    """Raised when a spider is misconfigured by a developer"""
+class SpiderInitializationError(KingfisherScrapyError):
+    """Base class for exceptions that occur before a spider is initialized"""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        logger.error('%s\n%s', str(self), traceback.format_stack()[-2].rstrip())
```
yolile, line 31: I'm wondering if we need to print the traceback too, as these are errors that we control so I'm not sure if the user needs to be aware of where the exception was raised. I guess there is no harm in printing the traceback but maybe they could be too noisy for a regular user and may indicate that there is an error with the code (but we know there is no error with the code but with the use). For example when not using the `sample` argument correctly

````
2021-11-24 16:01:39 [kingfisher_scrapy.exceptions] ERROR: spider argument `sample`: invalid integer value: 'a'
  File "kingfisher-collect/kingfisher_scrapy/base_spider.py", line 182, in from_crawler
    raise SpiderArgumentError(f'spider argument `sample`: invalid integer value: {spider.sample!r}')

PR #873, opened 2021-11-30
spiders(kosovo): update the spider with the OCDS URL and logic

yolile:
ref CRM-7714#note-34

### DIFF
--- a/kingfisher_scrapy/spiders/kosovo.py
+++ b/kingfisher_scrapy/spiders/kosovo.py
@@ -10,9 +10,11 @@ class Kosovo(SimpleSpider):
       Public Procurement Regulatory Commission
     Spider arguments
       from_date
-        Download only data from this time onward (YYYY-MM-DD format). Defaults to '2000-01-01'.
+        Download only data with a tender period end date from this time onward (YYYY-MM-DD format).
+        Defaults to '2000-01-01'.
       until_date
-        Download only data until this time (YYYY-MM-DD format). Defaults to now.
+        Download only data with a tender period end date until this time (YYYY-MM-DD format). Defaults to now.
+        Note: This date is non-inclusive.
     API documentation
       https://ocdskrpp.rks-gov.net/Help
     """
--- a/kingfisher_scrapy/spiders/kosovo.py
+++ b/kingfisher_scrapy/spiders/kosovo.py
@@ -1,5 +1,7 @@
+from datetime import timedelta
+
+from build.lib.kingfisher_scrapy.util import parameters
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/spiders/kosovo.py
```diff
@@ -8,37 +10,36 @@ class Kosovo(SimpleSpider):
...
-        for stage in stages:
-            yield self.build_request(
-                url.format(stage),
-                headers=headers,
-                formatter=components(-1)
-            )
+        # The API is slow even with short periods, so we request one day at a time.
+        delta = self.until_date - self.from_date
+        for days in reversed(range(delta.days + 1)):
+            start = self.from_date + timedelta(days=days-1)
```
jpmckinney, line 33: Can you check the API's logic? Typically I'd expect "from Nov 29 to 30" to mean "all of Nov 29 and all of Nov 30" (2 days), not "24:00:00 on Nov 29 to 23:59:00 on Nov 30" (1 day).

```suggestion
            start = self.from_date + timedelta(days=days - 1)
```

kingfisher_scrapy/spiders/kosovo.py
```diff
@@ -8,37 +10,36 @@ class Kosovo(SimpleSpider):
...
+            end = self.from_date + timedelta(days=days)
+            url = f'https://ocdskrpp.rks-gov.net/krppapi/tenderrelease?endDateFrom=' \
```
jpmckinney, line 35: I would have expected flake8 to complain about an f-string without a placeholder.

```suggestion
            url = 'https://ocdskrpp.rks-gov.net/krppapi/tenderrelease?endDateFrom=' \
```

kingfisher_scrapy/spiders/kosovo.py
yolile: Checked and, for example, https://ocdskrpp.rks-gov.net/krppapi/tenderrelease?endDateFrom=2021-11-26&endDateEnd=2021-11-27 only returns data from 2021-11-26 and similarly, https://ocdskrpp.rks-gov.net/krppapi/tenderrelease?endDateFrom=2021-11-25&endDateEnd=2021-11-26 only returns data from 2021-11-25 (the API filters is based on `tender/tenderPeriod/endDate`). So it is indeed just one day.

jpmckinney: Ok, please update the `until_date` docs to explain that this date is non-inclusive.

yolile: Done

jpmckinney (approved):

PR #891, opened 2021-12-30
update spider to make request per month

### LINKED ISSUE #882: `paraguay_dncp_*` get data by month
Currently, the pagination fails for pages larger than 1000. We can still get the data without errors if we always use enough filters, for example, requesting one month at a time.

cecicasco:
closes #882

### DIFF
--- a/kingfisher_scrapy/spiders/paraguay_dncp_base.py
+++ b/kingfisher_scrapy/spiders/paraguay_dncp_base.py
@@ -1,11 +1,12 @@
 import json
-from datetime import datetime
+from calendar import monthrange
+from datetime import datetime, timedelta
 
 import scrapy
 
 from kingfisher_scrapy.base_spider import SimpleSpider
 from kingfisher_scrapy.exceptions import AccessTokenError, MissingEnvVarError
-from kingfisher_scrapy.util import components, handle_http_error, parameters, replace_parameters
+from kingfisher_scrapy.util import components, date_range_by_month, handle_http_error, parameters, replace_parameters
 
 
 class ParaguayDNCPBase(SimpleSpider):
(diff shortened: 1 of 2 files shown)


yolile:
Thanks, @cecicasco. I made some changes to your code to ensure that we also respect the user's date parameters and also to re-request all the URLs that could be missed when requesting the access token. @jpmckinney, I would appreciate your input/review here as I think this is kind of complex.

kingfisher_scrapy/spiders/paraguay_dncp_base.py
```diff
@@ -45,17 +45,33 @@ def from_crawler(cls, crawler, *args, **kwargs):
...
+                # send duplicate requests when the token expired and in the continuation of last_requests saved.
+                dont_filter=True,
+                callback=self.parse_pages
+            )
+
+    def urls_builder(self):
+        days_interval = 30
+        # ElasticSearch doesn't allow searches sizes bigger than 10000.
+        # To avoid that, we request one month at the time.
+        days_to_iterate = timedelta(days=days_interval)
```
jpmckinney, line 61: We can use the timedelta object for comparisons, instead of comparing to the int object.

```suggestion
        # ElasticSearch doesn't allow search sizes greater than 10000, so we request one month at the time.
        interval = timedelta(days=30)
```

kingfisher_scrapy/spiders/paraguay_dncp_base.py
```diff
@@ -45,17 +45,33 @@ def from_crawler(cls, crawler, *args, **kwargs):
...
+        # In reverse chronological order
+        while end_date > self.from_date:
+            if (end_date - start_date).days > days_interval:
+                start_date = (end_date - days_to_iterate).replace(hour=0, minute=0, second=0)
+            else:
+                start_date = self.from_date
+            url = f'{self.base_url}/search/processes?tipo_fecha=fecha_release&' \
+                  f'fecha_desde={start_date.strftime(self.date_format)}-04:00&' \
+                  f'fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=500'
+            end_date = (start_date - timedelta(1)).replace(hour=23, minute=59, second=59)
```
jpmckinney, line 73: Use the keyword argument for clarity.

```suggestion
            end_date = (start_date - timedelta(days=1)).replace(hour=23, minute=59, second=59)
```

kingfisher_scrapy/spiders/paraguay_dncp_base.py
```diff
@@ -45,17 +45,33 @@ def from_crawler(cls, crawler, *args, **kwargs):
...
+        start_date = self.from_date
+        # In reverse chronological order
+        while end_date > self.from_date:
+            if (end_date - start_date).days > days_interval:
+                start_date = (end_date - days_to_iterate).replace(hour=0, minute=0, second=0)
+            else:
+                start_date = self.from_date
+            url = f'{self.base_url}/search/processes?tipo_fecha=fecha_release&' \
+                  f'fecha_desde={start_date.strftime(self.date_format)}-04:00&' \
+                  f'fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=500'
```
jpmckinney, line 72: Why is `500` chosen for `items_per_page`?

kingfisher_scrapy/spiders/paraguay_dncp_base.py
jpmckinney, line 73: I'm not clear why we replace the hours, minutes, seconds. Can't we do:

```
start_date = end_date - interval
...
end_date = start_date - timedelta(seconds=1)
```

I'm assuming that the API treats both fecha_desde and fecha_hasta as inclusive.

kingfisher_scrapy/spiders/paraguay_dncp_base.py
yolile: The default is 10, with 500 we are going to request fewer pages

jpmckinney: Is there a maximum? Would 1000 work? The comment says ElasticSearch limits to 10000, but if we can do 10000 per page that's probably even better.

kingfisher_scrapy/spiders/paraguay_dncp_base.py
yolile: Oh yes, you are right, substracting one second will do the trick as well!

kingfisher_scrapy/spiders/paraguay_dncp_base.py
yolile: 10.000 will work and we will always get only one page, so yes, we can use 10.000. The only trade-off (?) is that we will request each OCID faster now, as we are going to have the complete list of OCIDs with only one request. But I guess if we start to get 5xx errors from the API  we can use a greater `donwload_delay.

jpmckinney (approved):

PR #892, opened 2021-12-30
Handle first element of array is not a record file

### LINKED ISSUE #838: `mexico_quien_es_quien_records`: mixed data_types
https://api.quienesquien.wiki/v3/record returns a `data` array where the first entry is a record package but all the other entries are just records, without a package, which causes the `RootPathMiddleware` to fail.

cecicasco:
closes #838

### DIFF
--- a/kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
+++ b/kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
@@ -1,8 +1,8 @@
 from json import dumps
 
-
 from kingfisher_scrapy.spiders.mexico_quien_es_quien_base import MexicoQuienEsQuienBase
 from kingfisher_scrapy.util import handle_http_error
+
 
 class MexicoQuienEsQuienRecords(MexicoQuienEsQuienBase):
     """
--- a/kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
+++ b/kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
@@ -1,4 +1,4 @@
-from json import dumps
+import json
 
 from kingfisher_scrapy.spiders.mexico_quien_es_quien_base import MexicoQuienEsQuienBase
 from kingfisher_scrapy.util import handle_http_error
@@ -19,9 +19,6 @@ class MexicoQuienEsQuienRecords(MexicoQuienEsQuienBase):
     """
     name = 'mexico_quien_es_quien_records'
 
-    # BaseSpider
-    root_path = None  # The parse method already transforms the output with a record package at the root.
-
     # SimpleSpider
     data_type = 'record_package'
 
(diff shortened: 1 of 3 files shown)


yolile:
I changed the code a bit to use the record package metadata. @jpmckinney what do you think?

kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
```diff
@@ -1,19 +1,45 @@
...
     API documentation
       https://qqwapi-elastic.readthedocs.io/es/latest/
     Swagger API documentation
       https://api.quienesquien.wiki/v3/docs/
     """
     name = 'mexico_quien_es_quien_records'
 
+    # BaseSpider
+    root_path = None  # The parse method already transforms the output with a record package at the root.
+
```
jpmckinney, line 24: Instead of overriding in the subclass, move `root_path` out of the superclass into MexicoQuienEsQuienReleases

kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
```diff
@@ -1,19 +1,45 @@
+from json import dumps
```
jpmckinney, line 1: I prefer to just import `json`. This makes it easier to search for `json` methods (as we sometimes replace the json library).

kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
```diff
@@ -1,19 +1,45 @@
...
     # IndexSpider
     base_url = 'https://api.quienesquien.wiki/v3/record'
+    limit = 100  # Decrease the limit so the output file is not too big.
+
+    @handle_http_error
+    def parse(self, response):
+        data = response.json()
+        # The first entry of the array is a record package with 'records' as an object. The remaining entries
+        # are records. We use the package metadata to wrap all the records into a single record package.
+        if 'records' in data['data'][0]:
```
jpmckinney, line 37: In what case is `records` not in the first entry?

kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
```diff
@@ -1,19 +1,45 @@
...
+            package = data['data'][0].copy()
+            package['uri'] = None
+            package['publishedDate'] = None
+            package['records'] = [data['data'][0]['records']]
+            for record in data['data'][1:]:
+                package['records'].append(record)
```
jpmckinney, line 43: I think this is the same as:
```suggestion
            package['records'].extend(data['data'][1:])
```

yolile: ah yes, but without removing `package['records'] = [data['data'][0]['records']]` (otherwise we will loose the first record)

jpmckinney: Isn't the first record already at `package['records']` since it was `.copy()`-ed?

yolile: Nop, because the original 'records' key contains a record object, not an array

jpmckinney: Oh, weird, okay then! It can be `package['records'] = [package['records']]` before the `extend()`, then.

kingfisher_scrapy/spiders/mexico_quien_es_quien_records.py
yolile: Good question, I think never, and if they fix the issue in the future we will need to update the spider anyway so I will remove the check

jpmckinney (approved):
Very clean!

PR #900, opened 2022-01-17
database_store_extension: support compiling from list of releases from records

yolile:
This change is needed for Chile. Their records endpoint returns records with compile releases. However, some of them are incomplete, for example, https://apis.mercadopublico.cl/OCDS/data/record/4019-414-AG21
But they return the complete releases within the releases array, so we can compile the current data we have by using the release list from the records (we could also just use their releases endpoint instead, but if we do so, we will need to download all their data again)


yolile:
@jpmckinney, did you mean to approve this PR after your commit, or is there something else to change?

jpmckinney (approved):

PR #902, opened 2022-01-19
spiders: paraguay_dncp: request active/complete tender and planned ones separately

### LINKED ISSUE #882: `paraguay_dncp_*` get data by month

yolile:
closes #882

### DIFF
--- a/kingfisher_scrapy/spiders/paraguay_dncp_base.py
+++ b/kingfisher_scrapy/spiders/paraguay_dncp_base.py
@@ -56,7 +56,7 @@ class ParaguayDNCPBase(SimpleSpider):
 
     def urls_builder(self):
         # ElasticSearch doesn't allow search sizes greater than 10000, so we request half-month at the time.
-        interval = timedelta(days=15)
+        interval = timedelta(days=30)
         end_date = self.until_date
         # In reverse chronological order
         while end_date > self.from_date:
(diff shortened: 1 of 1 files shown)


kingfisher_scrapy/spiders/paraguay_dncp_base.py
```diff
@@ -65,11 +65,16 @@ def urls_builder(self):
...
-                  f'fecha_desde={start_date.strftime(self.date_format)}-04:00&' \
-                  f'fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=10000'
+            # We request active/complete tenders and planned ones separately to ensure we don't exceed the 10000
+            # results per request limit.
+            url_base = f'{self.base_url}/search/processes?fecha_desde={start_date.strftime(self.date_format)}-04:00' \
+                       f'&fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=10000 '
+            # We request the active or successful tenders by using the "publicacion_llamado" filter.
+            url = f'{url_base}&tipo_fecha=publicacion_llamado'
+            # And the planned ones with the "fecha_release" and tender.id="planned" filters.
+            url_planning = f'{url_base}&tender.id="planned"&tipo_fecha=fecha_release'
```
jpmckinney, line 75: Don't `"` need to be escaped in URLs?

kingfisher_scrapy/spiders/paraguay_dncp_base.py
```diff
@@ -65,11 +65,16 @@ def urls_builder(self):
...
                 start_date = end_date - interval
-            url = f'{self.base_url}/search/processes?tipo_fecha=fecha_release&' \
-                  f'fecha_desde={start_date.strftime(self.date_format)}-04:00&' \
-                  f'fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=10000'
+            # We request active/complete tenders and planned ones separately to ensure we don't exceed the 10000
+            # results per request limit.
+            url_base = f'{self.base_url}/search/processes?fecha_desde={start_date.strftime(self.date_format)}-04:00' \
+                       f'&fecha_hasta={end_date.strftime(self.date_format)}-04:00&items_per_page=10000 '
+            # We request the active or successful tenders by using the "publicacion_llamado" filter.
+            url = f'{url_base}&tipo_fecha=publicacion_llamado'
```
jpmckinney, line 73: Should we name this `url_tender` to clarify the difference from `url_planning` ?

yolile: Sure, although that URL will contain planning information as well, but only the ones which have a tender already

kingfisher_scrapy/spiders/paraguay_dncp_base.py
yolile: Actually, there is no need to use " at all

jpmckinney (approved):

PR #903, opened 2022-01-20
chore: split base_spider into separate files

### LINKED ISSUE #845: Split base_spider.py into separate files
Under a `kingfisher_scrapy.base_spiders` module.

yolile:
closes #845


docs/contributing/base_spiders/index.rst
```diff
@@ -0,0 +1,11 @@
+Base Classes
```
yolile, line 1: @jpmckinney I added this new file as the index for the base spiders (and to reference it in the contributing index), but not sure if needed

docs/contributing/base_spiders/index.rst
```diff
@@ -0,0 +1,11 @@
+Base Spiders
+------------
+
+.. toctree::
+   base_spider.rst
```
jpmckinney, line 5: Typically need a newline before content.

```suggestion

   base_spider.rst
```

docs/contributing/base_spiders/index.rst
```diff
@@ -0,0 +1,11 @@
...
+   big_file_spider.rst
+   compressed_file_spider.rst
+   index_spider.rst
+   links_spider.rst
+   periodic_spider.rst
+   simple_spider.rst
```
jpmckinney, line 11: Let's put these in class hierarchy order, then alphabetical order. So SimpleSpider would be after either BaseSpider or CompressedFileSpider (which also inherits from BaseSpider).

docs/contributing/base_spiders/index.rst
```diff
@@ -0,0 +1,11 @@
+Base Spiders
+------------
```
jpmckinney, line 2:
```suggestion
Base Spider Classes
-------------------
```

kingfisher_scrapy/base_spiders/base_spider.py
```diff
@@ -0,0 +1,325 @@
+import codecs
+from datetime import datetime
+
+import scrapy
+
+from kingfisher_scrapy.exceptions import IncoherentConfigurationError, SpiderArgumentError
+from kingfisher_scrapy.items import File, FileError, FileItem
+from kingfisher_scrapy.util import add_path_components, add_query_string
+
+browser_user_agent = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36'  # noqa: E501
```
jpmckinney, line 10: This isn't used here. Maybe move to util.py?

jpmckinney (commented):
I suggest importing all the classes into base_spiders/__init__.py – this means we can still import them as before (well, base_spiders instead of base_spider), and the import line is less verbose/duplicative. That's how I did it for extensions/__init__.py.

kingfisher_scrapy/base_spiders/base_spider.py
yolile: Yes, sure, I don't remember why we put it here in the first place

jpmckinney: I did 🙇‍♂️ but I probably wasn't thinking about it: https://github.com/open-contracting/kingfisher-collect/commit/f6c7dcb6b97ba5e2ee843869795fee6bc9bf3095

jpmckinney (approved):

jpmckinney (approved):

PR #908, opened 2022-01-28
docs: add note about checking and reporting the publication format

yolile:


docs/contributing/index.rst
```diff
@@ -11,6 +11,10 @@ Learn the data source's access methods
 
 Read its API documentation or bulk download documentation. Navigate the API, in your browser or with ``curl``. Inspect its responses, to determine where the OCDS data is located, and whether it includes information like pagination links, total pages or total results.
 
+.. note::
+
+   When doing this, check if the release or record packages are at the top-level of the JSON data. If not, please inform the analyst so this can be reported to the publisher.
```
jpmckinney, line 16:
```suggestion
   Please inform the helpdesk analyst of the following, so that it can be reported as feedback to the publisher:
   
   -  If there is no documentation about access methods
   -  If the release package or record package is *not* at the top-level of the JSON data
```

jpmckinney (commented):
Under "Write the spider", we can add after "After writing the spider, add a docstring for :ref:`spider metadata<spider-metadata>`." to be:

```
.. note::

   If you encountered any challenges, make a note in the docstring, and inform the helpdesk analyst, so that it can be reported as feedback to the publisher. Examples:

   -  Requests sometimes fail (e.g. timeout or error), but succeed on retry.
   -  Some requests always fail (e.g. for a specific date). 
   -  Error responses are returned with an HTTP 200 status code, instead of a status code in the range 400-599.
   -  The JSON data is encoded as ISO-8859-1, instead of UTF-8 per `RFC 8259 <https://datatracker.ietf.org/doc/html/rfc8259#section-8.1>`__.
   -  The number of results is limited to 10,000.
```

jpmckinney (approved):

PR #915, opened 2022-03-31
peru_compras: replace newline characters

### LINKED ISSUE #909: peru_compras: Contains literal LF character `\n` within string, which is invalid JSON
A newline character within a string needs to be escaped (`\\n`).

However, newlines can occur both within strings and at ends of lines, so unless we write a custom parser, it's not easy to escape only the newline characters within strings.

I guess we can replace `\n` with a single space.

Ravf95:
Closes #909

### DIFF
--- a/kingfisher_scrapy/spiders/peru_compras.py
+++ b/kingfisher_scrapy/spiders/peru_compras.py
@@ -1,3 +1,5 @@
+import json
+
 import scrapy
 
 from kingfisher_scrapy.base_spiders import SimpleSpider
@@ -49,3 +51,10 @@ class PeruCompras(SimpleSpider):
                     f'?pAcuerdo={framework_id}&pFechaIni={from_date}&pFechaFin={until_date}',
                     formatter=parameters('pAcuerdo')
                 )
+
+    @handle_http_error
+    def parse(self, response):
+        # Replace newline characters within strings with a space to prevent invalid JSON string
+        data = response.text.replace('\n', ' ')
+        response = response.replace(body=json.dumps(data))
+        yield self.build_file_from_response(response, data_type=self.data_type)


yolile (approved):

kingfisher_scrapy/spiders/peru_compras.py
```diff
@@ -49,3 +49,11 @@ def parse_list(self, response):
                     f'?pAcuerdo={framework_id}&pFechaIni={from_date}&pFechaFin={until_date}',
                     formatter=parameters('pAcuerdo')
                 )
+
+    @handle_http_error
+    def parse(self, response):
+        data = response.text
+        # Replace CR LF characters with spaces to prevent invalid JSON errors within strings
+        data = data.replace('\r', ' ').replace('\n', ' ')
```
jpmckinney, line 57: I haven't found any occurrences of the `\r` (CR) character. I've only found `&#x0D;`, which is its HTML encoding.

We can simplify this and only replace `\n`.

jpmckinney: Aha, yes, with `cat -e` I see the `^M` (CR) characters.

Ravf95: Maybe we can simplify with a single replace of \r\n

jpmckinney: In some files like pAcuerdo-232.json the `\n` occurs within the JSON string, without a preceding `\r`.

Ravf95: I understand, we can keep it for these cases

jpmckinney: It's fine for `\r` to appear outside strings (where it's valid). It's only a problem when it's unescaped within a string - but I think `\n` is the only character that is unescaped within strings.

PR #919, opened 2022-04-05
united_kingdom_contracts_finder: update to download record packages

### LINKED ISSUE #917: uk_contracts_finder: Add records spider
Please can we add a spider for the Contracts Finder records endpoint? In CRM-7756, the publisher clarified that historic releases are available via this endpoint, which are not available via the search endpoint.

The endpoint requires an OCID as a parameter. As far as I can see, the only way to get the OCID is via the search endpoint.

https://www.contractsfinder.service.gov.uk/apidocumentation/Notices/1/GET-Published-OCDS-Record
https://www.contractsfinder.service.gov.uk/apidocumentation/Notices/1/GET-Published-Notice-OCDS-Search

https://www.contractsfinder.service.gov.uk/Published/Notice/records/ocds-b5fd17-f6c03d3b-ba66-4b78-876a-8f3627330d36.json

**Link to CRM issue**
CRM-7756

**Priority**
As soon as possible - the publisher is actively engaging with the helpdesk about issues to do with linking releases.

**Data structure**
The data is structurally valid: [https://standard.open-contracting.org/review/data/dbad0031-8fbe-4253-b7a6-ba60e7b944e7](https://standard.open-contracting.org/review/data/dbad0031-8fbe-4253-b7a6-ba60e7b944e7)

**Publication format**
Record package

Ravf95:
Closes #917

### DIFF
--- a/kingfisher_scrapy/spiders/united_kingdom_contracts_finder.py
+++ b/kingfisher_scrapy/spiders/united_kingdom_contracts_finder.py
@@ -14,10 +14,9 @@ class UnitedKingdomContractsFinder(IndexSpider):
 
     # BaseSpider
     ocds_version = '1.0'  # uses deprecated fields
-    root_path = 'results.item'
 
     # SimpleSpider
-    data_type = 'release_package'
+    data_type = 'record_package'
 
     # IndexSpider
     total_pages_pointer = '/maxPage'
@@ -30,7 +29,11 @@ class UnitedKingdomContractsFinder(IndexSpider):
 
     def parse(self, response, **kwargs):
         if self.is_http_success(response):
-            yield from super().parse(response)
+            for result in response.json()['results']:
+                for release in result['releases']:
+                    ocid = release["ocid"]
+                    url = f'https://www.contractsfinder.service.gov.uk/Published/OCDS/Record/{ocid}'
+                    yield scrapy.Request(url, meta={'file_name': f'{ocid}.json'}, callback=super().parse)
         else:
             request = response.request.copy()
             wait_time = int(response.headers.get('Retry-After', 1))


kingfisher_scrapy/spiders/united_kingdom_contracts_finder.py
```diff
@@ -30,7 +29,11 @@ def start_requests(self):
 
     def parse(self, response, **kwargs):
         if self.is_http_success(response):
-            yield from super().parse(response)
+            for result in response.json()['results']:
+                for release in result['releases']:
+                    ocid = release["ocid"]
+                    url = f'https://www.contractsfinder.service.gov.uk/Published/OCDS/Record/{ocid}'
+                    yield scrapy.Request(url, meta={'file_name': f'{ocid}.json'}, callback=super().parse)
```
yolile, line 36: I suggest you keep this method as it is now and create a new one, for example, `build_urls` (you will need to set the `parse_list_callback` class variable to 'build_urls') and you will need to check for `self.is_http_success(response)` in this new method as well, in case the request fails here. And per https://github.com/open-contracting/kingfisher-collect/issues/917#issuecomment-1089394438 we will need to have two spiders, one for the releases and another one for the records. So you might want to create a base class.
And I assume by iterating the search endpoint we will have some duplicated ocids, so maybe we can implement a logic to avoid that or we can let scrapy drop the duplicated requests. @jpmckinney what do you think?

yolile (changes requested):

kingfisher_scrapy/spiders/united_kingdom_contracts_finder.py
jpmckinney: Let Scrapy drop the duplicate requests.

Ravf95: Thanks, I suggest we can use UnitedKingdomContractsFinder as a base class for the new spider (keeping parse list), and for the new one, we can implement a logic to avoid/replace the duplicated files.

Ravf95:
I close this PR to send a new one.

PR #920, opened 2022-04-06
add united_kingdom_contracts_finder_records spider

### LINKED ISSUE #917: uk_contracts_finder: Add records spider

Ravf95:
Closes #917

### DIFF
--- a/kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
+++ b/kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
@@ -15,4 +15,3 @@ class UnitedKingdomContractsFinderRecords(UnitedKingdomContractsFinderBase):
 
     # IndexSpider
     parse_list_callback = 'build_urls'
-
--- a/kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
+++ b/kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
@@ -15,9 +15,6 @@ class UnitedKingdomContractsFinderRecords(UnitedKingdomContractsFinderBase):
     # SimpleSpider
     data_type = 'record_package'
 
-    def parse(self, response):
-        for result in response.json()['results']:
-            for release in result['releases']:
-                ocid = release["ocid"]
-                url = f'https://www.contractsfinder.service.gov.uk/Published/OCDS/Record/{ocid}'
-                yield scrapy.Request(url, meta={'file_name': f'{ocid}.json'}, callback=super().parse)
+    # IndexSpider
+    parse_list_callback = 'build_urls'
+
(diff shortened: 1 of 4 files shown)


Ravf95:
@yolile @jpmckinney I checked the [isort documentation](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html) but I don't understand why the check fails

jpmckinney:
Just run `isort . --line-width 119` to fix it.

kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
```diff
@@ -0,0 +1,23 @@
...
+    API documentation
+      https://www.contractsfinder.service.gov.uk/apidocumentation/home
+    """
+    name = 'united_kingdom_contracts_finder_records'
+
+    # SimpleSpider
+    data_type = 'record_package'
+
+    def parse(self, response):
+        for result in response.json()['results']:
```
yolile, line 19: We need to check here if the response was successful and if not, apply the same existing logic. So, instead of overriding the parse method, I would suggest adding a new one that checks if the response was successful, build the URLs, and then callback the parse method. Same for the releases spider that also needs to build the URLs.

yolile (changes requested):

kingfisher_scrapy/spiders/united_kingdom_contracts_finder_releases.py
```diff
@@ -0,0 +1,17 @@
+from kingfisher_scrapy.spiders.united_kingdom_contracts_finder_base import UnitedKingdomContractsFinderBase
+
+
+class UnitedKingdomContractsFinderReleases(UnitedKingdomContractsFinderBase):
```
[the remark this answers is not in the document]
Ravf95: Done

kingfisher_scrapy/spiders/united_kingdom_contracts_finder_records.py
Ravf95: Done

yolile:
I've refactored the code a little bit to avoid some duplications. @jpmckinney not sure if we should do something else or different to improve this further

yolile (dismissed):

jpmckinney (approved):

PR #926, opened 2022-04-08
chore: Follow consistent approach for class attributes of publisher/system-specific base classes

jpmckinney:


yolile (dismissed):
I've added the comment for MexicoINAIBase and NigeriaPortalBase as well

jpmckinney:
Thanks! That reminds me there was another one for `EuropeanDynamicsBase`.

PR #927, opened 2022-04-20
nigeria_budeshi: Share logic with Kaduna State Budeshi spiders

jpmckinney:


yolile (approved):
I guess this is working, but I couldn't test it because the site returns 404 now. I found some contracts when searching for "Kaduna" at https://budeshi.ng/Home/projects so maybe everything is there now? In any case, I will merge this one and open a separate issue.

yolile:
Edit: it is not returning 404, but 200 with an empty response

PR #928, opened 2022-04-20
paraguay: Reduce duplication and fix bugs

jpmckinney:
- paraguay_hacienca: Was setting last_request scalar, but middleware expects last_requests list.
- Set access_token_scheduled_at on each attempt, not only on the first attempt.

### DIFF
--- a/kingfisher_scrapy/downloadermiddlewares.py
+++ b/kingfisher_scrapy/downloadermiddlewares.py
@@ -36,6 +36,8 @@ class ParaguayAuthMiddleware:
             requests_backlog = []
 
             def build_access_token_request(self):
+                self.access_token_scheduled_at = datetime.now()
+
                 return scrapy.Request("https://example.com")
     """
 
--- a/kingfisher_scrapy/downloadermiddlewares.py
+++ b/kingfisher_scrapy/downloadermiddlewares.py
@@ -21,6 +21,22 @@ class ParaguayAuthMiddleware:
     Tokens should be generated and assigned just before sending a request, but Scrapy does not provide any way to do
     this, which in turn means that sometimes we accidently send expired tokens. For now, the issue seems to be avoided
     by setting the number of concurrent requests to 1, at cost of download speed.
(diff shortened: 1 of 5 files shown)


kingfisher_scrapy/downloadermiddlewares.py
```diff
@@ -33,36 +51,36 @@ def from_crawler(cls, crawler):
...
+            spider.crawler.engine.close_spider(spider, 'access_token_request_failed')
             raise IgnoreRequest("Max attempts to get an access token reached. Stopping crawl...")
         request.headers['Authorization'] = spider.access_token
         if self._expires_soon(spider):
-            # SAVE the last request to continue after getting the token
-            spider.last_requests.append(request)
-            spider.logger.info('Saving request for after getting the token: %s', request.url)
-            # spider MUST implement the request_access_token method
-            return spider.request_access_token()
+            return spider.add_request_to_backlog_and_build_access_token_request(spider, request)
```
yolile, line 59:
```suggestion
            return self.add_request_to_backlog_and_build_access_token_request(spider, request)
```

kingfisher_scrapy/spiders/paraguay_dncp_records.py
```diff
@@ -24,6 +24,6 @@ class ParaguayDNCPRecords(ParaguayDNCPBase):
     # SimpleSpider
     data_type = 'record_package'
 
-    def get_files_to_download(self, content):
-        for record in content['records']:
-            yield f"{self.base_url}/ocds/record/{record['ocid']}"
+    def get_files_to_download(self, data):
+        for record in data['records']:
+            yield f"{self.url_prefix}/ocds/record/{record['ocid']}"
```
yolile, line 29:
```suggestion
            yield f"{self.url_prefix}ocds/record/{record['ocid']}"
```

yolile (changes requested):

yolile (approved):

PR #935, opened 2022-05-23
spiders: add ecuador_sercop_api

### LINKED ISSUE #934: Ecuador SERCOP API
We already have a spider for Ecuador SERCOP that uses their bulk download option. However, due to the current Kingfisher Process issue on processing `FileItem` we are having issues collecting all Ecuador's data (see the errors in the Ecuador latest [log](https://collect.kingfisher.open-contracting.org/logs/kingfisher/ecuador_sercop/58a018eccfa311ec88920c9d92c523cb.log)). We will need to use the API instead to get all the data, as the API has a service to get a release package per ocid.
Note that we will need to rename `ecuador_sercop` to `ecuador_sercop_bulk`

https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/datos-abiertos/api

https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/datos-abiertos/api

**Priority**
anytime

**Data structure**
Valid OCDS 1.1 data

**Publication format**
A list of ocids and release packages

yolile:
closes #934


jpmckinney:
Noting a couple errors in Sentry.

yolile:
@jpmckinney I've implemented the changes

kingfisher_scrapy/commands/checkall.py
```diff
@@ -76,7 +76,7 @@ def check(self):
         basename = os.path.splitext(os.path.basename(self.module.__file__))[0]
         expected_basename = re.sub(word_boundary_re, '_', class_name).lower()
 
-        if basename != expected_basename and class_name != 'PakistanPPRAAPI':
+        if basename != expected_basename and class_name != 'PakistanPPRAAPI' and class_name != 'EcuadorSERCOPAPI':
```
jpmckinney, line 79:
```suggestion
        if basename != expected_basename and class_name not in ('PakistanPPRAAPI', 'EcuadorSERCOPAPI'):
```

kingfisher_scrapy/spiders/ecuador_sercop_api.py
```diff
@@ -0,0 +1,79 @@
...
+    parse_list_callback = 'parse_page'
+    start_requests_callback = 'parse_list'
+    total_pages_pointer = '/pages'
+
+    # Local
+    url_prefix = 'https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/api/'
+
+    # PeriodicSpider
+    formatter = staticmethod(components(-1))
+    pattern = f'{url_prefix}search_ocds'+'?year={0}'
```
jpmckinney, line 44:
```suggestion
    # Local
    url_prefix = 'https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/api/'

    # PeriodicSpider
    formatter = staticmethod(components(-1))
    pattern = f'{url_prefix}search_ocds?year={{0}}'
    start_requests_callback = 'parse_list'

    # IndexSpider
    total_pages_pointer = '/pages'
    parse_list_callback = 'parse_page'
```

kingfisher_scrapy/spiders/ecuador_sercop_api.py
```diff
@@ -0,0 +1,79 @@
...
+    def parse(self, response, **kwargs):
+        if self.is_http_success(response):
+            yield from super().parse(response)
+        else:
+            yield self.build_retry_request_or_file_error(response)
+
+    def build_retry_request_or_file_error(self, response):
+        if response.status == 429:
+            request = response.request.copy()
+            wait_time = int(response.headers.get('retry-after', 1))
```
jpmckinney, line 70: I prefer the correct lettercase, unless it doesn't work for some reason.

Also, does the 429 response not always set the Retry-After header?

```suggestion
            wait_time = int(response.headers.get('Retry-After', 1))
```

kingfisher_scrapy/spiders/ecuador_sercop_api.py
```diff
@@ -0,0 +1,79 @@
...
+            yield from super().parse_list(response)
+        else:
+            yield self.build_retry_request_or_file_error(response)
+
+    def parse_page(self, response):
+        if self.is_http_success(response):
+            for data in response.json()['data']:
+                # Some ocids have a '/' character which cannot be in a file name.
+                yield self.build_request(f'{self.url_prefix}record?ocid={data["ocid"]}',
+                                         formatter=lambda file_name: parameters('ocid')(file_name).replace('/', ''))
```
jpmckinney, line 57: Let's preserve the token boundaries in the OCID by replacing / with _
```suggestion
                                         formatter=lambda url: parameters('ocid')(url).replace('/', '_'))
```

kingfisher_scrapy/spiders/ecuador_sercop_api.py
```diff
@@ -0,0 +1,79 @@
...
+        https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/datos-abiertos/api
+    Bulk download documentation
+      https://datosabiertos.compraspublicas.gob.ec/PLATAFORMA/datos-abiertos
+    """
+    name = 'ecuador_sercop_api'
+    # The API returns HTTP error 429 after a number of requests, so we have our own retry logic.
+    # We also reduce the number of concurrent requests to avoid too many failures.
+    custom_settings = {
+        'CONCURRENT_REQUESTS': 2,
+        'RETRY_HTTP_CODES': [],
```
jpmckinney, line 24: Why is this necessary? 

https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-http-codes

kingfisher_scrapy/spiders/ecuador_sercop_api.py
yolile: I tested to be sure, and yes, It works with the correct letter case too. 

> Also, does the 429 response not always set the Retry-After header?

So far, the response always has set the header, so yes, we can remove the default value.

jpmckinney: Let's remove the default value. We should only set defaults when a value is *known* to be missing sometimes.

kingfisher_scrapy/spiders/ecuador_sercop_api.py
```diff
@@ -0,0 +1,79 @@
...
+    def parse(self, response, **kwargs):
+        if self.is_http_success(response):
+            yield from super().parse(response)
+        else:
+            yield self.build_retry_request_or_file_error(response)
+
+    def build_retry_request_or_file_error(self, response):
+        if response.status == 429:
+            request = response.request.copy()
+            wait_time = int(response.headers.get('Retry-After'))
```
jpmckinney, line 70: Otherwise it returns `None`, and `int(None)` raises a TypeError which is less clear than a KeyError.

```suggestion
            wait_time = int(response.headers['Retry-After'])
```

yolile:
This version of the spider [has been running](https://collect.kingfisher.open-contracting.org/logs/kingfisher/ecuador_sercop_api/5f116afcdba411ec88920c9d92c523cb.log) for 2 days now. I'm not sure if just because of the number of 429 errors or because the duplicate URLs are repeatedly retrying, should we limit the number of retries per URL?. If yes,  I'm not sure how many times we should retry, as a very unlucky URL could never have time to be retried.

jpmckinney:
Yes, as mentioned in https://github.com/open-contracting/kingfisher-collect/issues/917#issuecomment-1104277470, we should always have `max_attempts` or `max_retries`.

yolile:
I've moved the retry logic to `BaseSpider` and updated UK, Portugal, and Ecuador to use that method instead

yolile:
@jpmckinney If you are ok with this approach, I will add a test for it.

jpmckinney:
I made a commit to use `handle_http_error` instead.

jpmckinney (dismissed):
I split up the tests, because testing multiple branches in the same test becomes hard to read.

yolile:
Sorry, I need your approval again as I made a small commit https://github.com/open-contracting/kingfisher-collect/pull/935/commits/453eccef64a7e29a536209d9c161e93c49f853aa to remove an unnecessary check

PR #940, opened 2022-06-01
ecuador_sercop_bulk: avoid losing data in kingfisher process

yolile:
Following from #940
The API is too slow, so this is a workaround to get all the data until kingfisher process is able to process a lot of file items simultaneously.

### DIFF
--- a/kingfisher_scrapy/spiders/ecuador_sercop_bulk.py
+++ b/kingfisher_scrapy/spiders/ecuador_sercop_bulk.py
@@ -1,5 +1,8 @@
+import ijson
+
+from kingfisher_scrapy import util
 from kingfisher_scrapy.base_spiders import CompressedFileSpider, PeriodicSpider
-from kingfisher_scrapy.util import components
+from kingfisher_scrapy.util import components, handle_http_error
 
 
 class EcuadorSERCOPBulk(CompressedFileSpider, PeriodicSpider):
@@ -21,7 +24,6 @@ class EcuadorSERCOPBulk(CompressedFileSpider, PeriodicSpider):
     # BaseSpider
     date_format = 'year-month'
     default_from_date = '2015-01'
-    root_path = 'item'
 
     # SimpleSpider
     data_type = 'release_package'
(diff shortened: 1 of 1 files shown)


jpmckinney:
I'm not sure why https://github.com/open-contracting/kingfisher-process/issues/353 is linked. The data loss is due to the helpdesk server running the old version of Kingfisher Process. There is no data loss in the new version.

kingfisher_scrapy/spiders/ecuador_sercop_bulk.py
```diff
@@ -34,3 +36,20 @@ class EcuadorSERCOPBulk(CompressedFileSpider, PeriodicSpider):
...
         return super().build_request(url, formatter, **meta)
+
+    @handle_http_error
+    def parse(self, response):
+        item = next(super().parse(response))
+        data = item['data']
+        package = {}
+        # The file is a JSON array of release packages that we rearrange in a single package to avoid losing data
+        # in Kingfisher Process https://github.com/open-contracting/kingfisher-process/issues/353.
+        for number, obj in enumerate(util.transcode(self, ijson.items, data, 'item'), 1):
```
jpmckinney, line 47: We don't need to transcode here as Ecuador uses UTF-8. We have `transcode` elsewhere because we can't guarantee the encoding across all sources.

yolile:
Changed

jpmckinney (commented):
This might be better as a spider middleware. I think some other sources that use `root_path` (with `item` somewhere in the root path) suffer from the same design issue, like OpenOpps.

Although Process v2 can handle it, sending millions more RabbitMQ messages than necessary is still slower than performing a bit of aggregation in Collect.

yolile:
> As I think about it more, maybe if 'item' in root_path.split('.'):, and if the data_type is release_package or record_package, then RootPathMiddleware should always combine the packages.

Yes, I thought about updating `RootPathMiddleware` so we apply the same solution for the cases you mentioned. I can change the middleware to do that if you are happy with that.

yolile:
I will close this PR and open an issue to discuss

PR #943, opened 2022-06-01
spidermiddlewares: support resizing record packages

### LINKED ISSUE #941: ResizePackageMiddleware: Add support for record packages
It can check the `data_type` to know whether to use `releases` or `records`.

This middleware runs when `resize_package = True`, which is the default for BigFileSpider and can be set for CompressedFileSpider.

BigFileSpider is explicitly only for release packages, but CompressedFileSpider can be used for records (like ChileCompraBulk).

Right now, the only CompressedFileSpider to use `resize_package` is DominicanRepublicBulk, which is for release packages, so there is no problem with an existing spider.

That said, this should be a simple change. We can also then edit BigFileSpider to allow its use with record packages.

yolile:
closes #941

### DIFF
--- a/kingfisher_scrapy/base_spiders/big_file_spider.py
+++ b/kingfisher_scrapy/base_spiders/big_file_spider.py
@@ -34,7 +34,7 @@ class BigFileSpider(SimpleSpider):
     def from_crawler(cls, crawler, *args, **kwargs):
         spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
 
-        if not getattr(spider, 'data_type') or spider.data_type not in ('release_package', 'record_package'):
+        if spider.data_type not in ('release_package', 'record_package'):
             raise IncoherentConfigurationError(
                 f"data_type must be 'release_package' or 'record_package', not {spider.data_type!r}")
 
--- a/kingfisher_scrapy/base_spiders/big_file_spider.py
+++ b/kingfisher_scrapy/base_spiders/big_file_spider.py
@@ -34,8 +34,9 @@ class BigFileSpider(SimpleSpider):
     def from_crawler(cls, crawler, *args, **kwargs):
         spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
 
-        if not spider.data_type or spider.data_type not in ('release_package', 'record_package'):
-            raise IncoherentConfigurationError(f'{spider.data_type} data_type is not compatible with BigFileSpider')
+        if spider.data_type not in ('release_package', 'record_package'):
+            raise IncoherentConfigurationError(
+                f"data_type must be 'release_package' or 'record_package', not {spider.data_type!r}")
 
         return spider
 
--- a/kingfisher_scrapy/base_spiders/big_file_spider.py
+++ b/kingfisher_scrapy/base_spiders/big_file_spider.py
@@ -34,7 +34,7 @@ class BigFileSpider(SimpleSpider):
     def from_crawler(cls, crawler, *args, **kwargs):
         spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
 
-        if spider.data_type not in ('release_package', 'record_package'):
+        if not getattr(spider, 'data_type') or spider.data_type not in ('release_package', 'record_package'):
             raise IncoherentConfigurationError(
                 f"data_type must be 'release_package' or 'record_package', not {spider.data_type!r}")
 
--- a/kingfisher_scrapy/base_spiders/big_file_spider.py
+++ b/kingfisher_scrapy/base_spiders/big_file_spider.py
@@ -1,4 +1,5 @@
 from kingfisher_scrapy.base_spiders import SimpleSpider
+from kingfisher_scrapy.exceptions import IncoherentConfigurationError
 from kingfisher_scrapy.util import handle_http_error
 
 
@@ -28,6 +29,15 @@ class BigFileSpider(SimpleSpider):
     """
 
     resize_package = True
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
+
+        if not spider.data_type or spider.data_type not in ('release_package', 'record_package'):
+            raise IncoherentConfigurationError(f'{spider.data_type} data_type is not compatible with BigFileSpider')
+
+        return spider
 
     @handle_http_error
     def parse(self, response):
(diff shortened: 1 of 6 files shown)


kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -161,16 +161,19 @@ def process_spider_output(self, response, result, spider):
...
             else:
                 size = 100
-
-            package = self._get_package_metadata(spider, data['package'], 'releases', item['data_type'])
-            iterable = util.transcode(spider, ijson.items, data['data'], 'releases.item')
+            if spider.data_type == 'release_package':
+                package_type = 'releases'
+            else:
+                package_type = 'records'
+            package = self._get_package_metadata(spider, data['package'], package_type, item['data_type'])
```
jpmckinney, line 168: Newline to separate code into logic blocks.

```suggestion

            package = self._get_package_metadata(spider, data['package'], package_type, item['data_type'])
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -161,16 +161,19 @@ def process_spider_output(self, response, result, spider):
                 size = spider.sample
             else:
                 size = 100
-
-            package = self._get_package_metadata(spider, data['package'], 'releases', item['data_type'])
-            iterable = util.transcode(spider, ijson.items, data['data'], 'releases.item')
+            if spider.data_type == 'release_package':
+                package_type = 'releases'
```
jpmckinney, line 165: Change to `key`, as was done in the other PR.

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -161,16 +161,19 @@ def process_spider_output(self, response, result, spider):
...
+            else:
+                package_type = 'records'
+            package = self._get_package_metadata(spider, data['package'], package_type, item['data_type'])
+            iterable = util.transcode(spider, ijson.items, data['data'], f'{package_type}.item')
             # We yield release packages containing a maximum of 100 releases.
```
jpmckinney, line 170:
```suggestion
            # We yield packages containing a maximum of 100 items.
```

jpmckinney (commented):
We can also update BigFileSpider to allow either `data_type = 'release_package'` or `data_type = 'record_package'`, and update its documentation. A `from_crawler` method can error if data_type is not one of these.

yolile:
> We can also update BigFileSpider to allow either `data_type = 'release_package'` or `data_type = 'record_package'`, and update its documentation. A `from_crawler` method can error if data_type is not one of these.

Done!

kingfisher_scrapy/base_spiders/big_file_spider.py
```diff
@@ -29,7 +30,16 @@ def start_requests(self):
 
     resize_package = True
 
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
+
+        if not spider.data_type or spider.data_type not in ('release_package', 'record_package'):
```
jpmckinney, line 37: This is simpler and has the same outcome.

```suggestion
        if spider.data_type not in ('release_package', 'record_package'):
```

kingfisher_scrapy/base_spiders/big_file_spider.py
```diff
@@ -29,7 +30,16 @@ def start_requests(self):
...
+            raise IncoherentConfigurationError(f'{spider.data_type} data_type is not compatible with BigFileSpider')
```
jpmckinney, line 38: Provide more relevant feedback.

```suggestion
            raise IncoherentConfigurationError(
                f"data_type must be 'release_package' or 'record_package', not {spider.data_type!r}")
```

kingfisher_scrapy/base_spiders/simple_spider.py
```diff
@@ -26,6 +26,8 @@ def start_requests(self):
                 yield scrapy.Request('https://example.com/api/package.json', meta={'file_name': 'all.json'})
     """
 
+    data_type = None
+
```
jpmckinney, line 30: We didn't set this, so that code that uses `self.data_type` will error. That way, we don't need to add checks all over the place to guarantee that `self.data_type is not None`. I would remove it.

tests/base_spiders/test_big_file_spider.py
```diff
@@ -3,16 +3,33 @@
...
+    # No IncoherentConfigurationError exception.
+    spider_with_crawler(BigFileSpider, data_type=data_type)
+
+
+@pytest.mark.parametrize('sample,len_items,len_releases,data_type', [
+    (None, 2, 100, 'release_package'), (5, 1, 5, 'release_package'), (5, 1, 5, 'record_package')
+])
+def test_parse_package(sample, len_items, len_releases, data_type):
+    spider = spider_with_crawler(spider_class=BigFileSpider, sample=sample, data_type=data_type)
+    key = f'{data_type[-8]}s'
```
jpmckinney, line 29: Put the `key` in the `parametrize` tuples. It's easier to read that way.

tests/test_spidermiddlewares.py
```diff
@@ -188,17 +188,18 @@ def test_add_package_middleware(data_type, data, root_path):
 
 @pytest.mark.parametrize('sample,len_items,len_releases', [(None, 2, 100), (5, 5, 5)])
 @pytest.mark.parametrize('encoding,character', [('utf-8', b'\xc3\x9a'), ('iso-8859-1', b'\xda')])
-def test_resize_package_middleware(sample, len_items, len_releases, encoding, character):
+@pytest.mark.parametrize('package_type', ['record', 'release'])
+def test_resize_package_middleware(sample, len_items, len_releases, encoding, character, package_type):
```
jpmckinney, line 192: Similarly, add each of `data_type` and `key` to the parametrize tuples, to make this easier to read. Reading code in order to understand the test's inputs takes extra time.

kingfisher_scrapy/base_spiders/big_file_spider.py
```diff
@@ -34,7 +34,7 @@ def start_requests(self):
     def from_crawler(cls, crawler, *args, **kwargs):
         spider = super(BigFileSpider, cls).from_crawler(crawler, *args, **kwargs)
 
-        if spider.data_type not in ('release_package', 'record_package'):
+        if not getattr(spider, 'data_type') or spider.data_type not in ('release_package', 'record_package'):
             raise IncoherentConfigurationError(
                 f"data_type must be 'release_package' or 'record_package', not {spider.data_type!r}")
```
jpmckinney, line 39: I'm pretty sure this line will error when it tries to access `spider.data_type` if `not getattr(spider, 'data_type')` is True. I don't think `not getattr(spider, 'data_type')` is needed. If the spider developer failed to set the data_type, the user will see something like "AttributeError: 'BigFileSpider' object has no attribute 'data_type'", which clearly identifies the problem.

yolile: Ops, you are right

jpmckinney (approved):

PR #944, opened 2022-06-02
RootPathMiddleware: group items in packages

### LINKED ISSUE #942: Limit the number of FileItem with spider middlewares
@jpmckinney 

From https://github.com/open-contracting/kingfisher-collect/pull/940#pullrequestreview-992316521:

> This might be better as a spider middleware. I think some other sources that use root_path (with item somewhere in the root path) suffer from the same design issue, like OpenOpps.
>
> Although Process v2 can handle it, sending millions more RabbitMQ messages than necessary is still slower than performing a bit of aggregation in Collect.

and https://github.com/open-contracting/kingfisher-collect/pull/940#issuecomment-1143909378:

> As I think about it more, maybe if 'item' in root_path.split('.'):, and if the data_type is release_package or record_package, then RootPathMiddleware should always combine the packages.

@yolile

> Yes, I thought about updating RootPathMiddleware so we apply the same solution for the cases you mentioned. I can change the middleware to do that if you are happy with that.

> We could do the same for line delimiter files, for example, Digiwhist, although we will need to limit the number of releases per package (for big files), similar to `ResizePackageMiddleware`. And we could do the same for releases and records, but adding them to an empty package first, as in `AddPackageMiddleware`

yolile:
closes #942

### DIFF
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -85,7 +85,18 @@ class RootPathMiddleware:
 
             if isinstance(data, (dict, list)):
                 data = util.json_dumps(data).encode()
+
+            if 'release' in item['data_type']:
+                key = 'releases'
+                data_type = 'release_package'
+            else:
+                key = 'records'
+                data_type = 'record_package'
+
             for number, obj in enumerate(util.transcode(spider, ijson.items, data, spider.root_path), 1):
+                # Avoid reading the rest of a large file, since the rest of the items will be dropped.
+                if spider.sample and number > spider.sample:
+                    break
 
                 # Two common issues in OCDS data are:
                 #
@@ -96,47 +107,22 @@ class RootPathMiddleware:
                 # written, the number of messages in RabbitMQ and the number of rows in PostgreSQL.
                 #
                 # We fix the packaging to reduce the overhead.
-                if 'item' in spider.root_path.split('.') and not package:
-                    if 'release' in item['data_type']:
-                        package_type = 'releases'
-                        data_type = 'release_package'
-                    else:
-                        package_type = 'records'
-                        data_type = 'record_package'
+                if 'item' in spider.root_path.split('.'):
                     # Assume that the `extensions` are the same for all packages.
-                    if 'package' in item['data_type']:
-                        package = obj.copy()
-                    else:
-                        package = {}
-                    package[package_type] = []
-
-                # Avoid reading the rest of a large file, since the rest of the items will be dropped.
-                if spider.sample and number > spider.sample:
-                    break
-
-                if isinstance(item, File):
-                    if package:
+                    if not package:
                         if 'package' in item['data_type']:
-                            package[package_type].extend(obj[package_type])
+                            package = obj.copy()
                         else:
-                            package[package_type].append(obj)
+                            package = {}
+                        package[key] = []
+
+                    if 'package' in item['data_type']:
+                        package[key].extend(obj[key])
                     else:
-                        item['data'] = obj
-                        yield item
+                        package[key].append(obj)
                 else:
-                    # If the input data was a JSON stream and the root path is to a JSON array, then this method will
-                    # need to yield multiple FileItems for each input FileItem. To do so, the input FileItem's number
-                    # is multiplied by the maximum length of the JSON array, to avoid duplicate numbers.
-                    #
-                    # If this is the case, then, on the spider, set a ``root_path_max_length`` class attribute to the
-                    # maximum length of the JSON array at the root path.
-                    #
-                    # Note that, to be stored by Kingfisher Process, the final number must be within PostgreSQL's
-                    # integer range.
-                    #
-                    # https://www.postgresql.org/docs/11/datatype-numeric.html
-                    yield spider.build_file_item((item['number'] - 1) * spider.root_path_max_length + number,
-                                                 obj, item)
+                    item['data'] = obj
+                    yield item
             if package:
                 item['data_type'] = data_type
                 item['data'] = package
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -113,7 +113,7 @@ class RootPathMiddleware:
                         if 'package' in item['data_type']:
                             package = obj.copy()
                         else:
-                            package = {}
+                            package = {'version': spider.ocds_version}
                         package[key] = []
 
                     if 'package' in item['data_type']:
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -64,7 +64,10 @@ class LineDelimitedMiddleware:
 
 class RootPathMiddleware:
     """
-    If the spider's ``root_path`` class attribute is non-empty, yields a FileItem for each object at that prefix.
+    If the spider's ``root_path`` class attribute is non-empty, then:
+     - If the file contains an array ('item' in ``root_path``), packages the releases or records in the same package
+       and yields a File
+     - If not, yields the object at ``root_path`` as a File
     Otherwise, yields the original item.
     """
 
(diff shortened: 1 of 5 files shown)


yolile:
@jpmckinney I will update the tests if you are ok with the approach

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
 
             data = item['data']
 
+            package = None
+
             if isinstance(data, (dict, list)):
                 data = util.json_dumps(data).encode()
-
```
jpmckinney, line 84: Let's leave this newline. The encoding of `data` is not part of the same logical group as the for-loop.

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
+                    else:
+                        package = {}
+                    package[package_type] = []
+
                 # Avoid reading the rest of a large file, since the rest of the items will be dropped.
                 if spider.sample and number > spider.sample:
-                    return
+                    break
 
                 if isinstance(item, File):
```
jpmckinney, line 110: Once we finish the refactor, we shouldn't need any logic around `root_path_max_length`, I think. If so, then we can reorganize the if-statements a bit, e.g.

```
if 'item' in spider.root_path.split('.'):
    if not package:
        # initialize package
    # update package
else:
    # ...
    yield item

File: kingfisher_scrapy/spidermiddlewares.py [hunk 3]
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
 
             data = item['data']
 
+            package = None
+
             if isinstance(data, (dict, list)):
                 data = util.json_dumps(data).encode()
-
             for number, obj in enumerate(util.transcode(spider, ijson.items, data, spider.root_path), 1):
+
```
jpmckinney, line 90: Avoid an extra newline inside a for-loop.

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
+                # If the file contains an array of either packages or non-packaged releases and records, we group them
+                # in a single package
```
jpmckinney, line 92:
```suggestion
                # Two common issues in OCDS data are:
                #
                # - Multiple releases or records, without a package
                # - Multiple packages in a single file (often with a single release, record or OCID per package)
                #
                # Yielding each release, record or package creates a lot of overhead in terms of the number of files
                # written, the number of messages in RabbitMQ and the number of rows in PostgreSQL.
                #
                # We fix the packaging to reduce the overhead.
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -64,7 +64,10 @@ def process_spider_output(self, response, result, spider):
 
 class RootPathMiddleware:
     """
-    If the spider's ``root_path`` class attribute is non-empty, yields a FileItem for each object at that prefix.
+    If the spider's ``root_path`` class attribute is non-empty, then:
+     - If the file contains an array ('item' in ``root_path``), packages the releases or records in the same package
+       and yields a File
+     - If not, yields the object at ``root_path`` as a File
     Otherwise, yields the original item.
```
jpmckinney, line 71:
```suggestion
    If the spider's ``root_path`` class attribute is non-empty, yields an item for the object(s) at the ``root_path``.
    Otherwise, yields the original item.

    If the objects (releases, records or packages) are within an array, they are combined into a single package.
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
+                        package_type = 'releases'
+                        data_type = 'release_package'
+                    else:
+                        package_type = 'records'
+                        data_type = 'record_package'
+                    if 'package' in item['data_type']:
+                        package = obj.copy()
+                    else:
+                        package = {}
+                    package[package_type] = []
```
jpmckinney, line 104: In the tests, please check that `copy` left the original list intact (for `extend` later). I'm not 100% sure that it does.

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
+                # If the file contains an array of either packages or non-packaged releases and records, we group them
+                # in a single package
+                if 'item' in spider.root_path.split('.') and not package:
+                    if 'release' in item['data_type']:
+                        package_type = 'releases'
+                        data_type = 'release_package'
+                    else:
+                        package_type = 'records'
+                        data_type = 'record_package'
+                    if 'package' in item['data_type']:
```
jpmckinney, line 100:
```suggestion
                    # Assume that the `extensions` are the same for all packages.
                    if 'package' in item['data_type']:
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
+                        package = obj.copy()
+                    else:
+                        package = {}
+                    package[package_type] = []
+
                 # Avoid reading the rest of a large file, since the rest of the items will be dropped.
                 if spider.sample and number > spider.sample:
-                    return
+                    break
```
jpmckinney, line 108: I think this if-statement can be at the start of the loop?

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -79,16 +82,40 @@ def process_spider_output(self, response, result, spider):
...
                 if isinstance(item, File):
-                    yield spider.build_file_item(number, obj, item)
+                    if package:
+                        if 'package' in item['data_type']:
+                            package[package_type].extend(obj[package_type])
+                        else:
+                            package[package_type].append(obj)
+                    else:
+                        item['data'] = obj
+                        yield item
```
jpmckinney, line 118: The original code always built a file item.

yolile: In the previous version, we yielded FileItems when the File:
- Had an array (now we are editing the same File, so there is no need to yield FileItems)
- Hadn't a release, record, or package at the top level. So still, no need to yield FileItems in that case, just an updated File

jpmckinney: Okay, makes sense to always have a File then.

yolile: There is still one case when we will yield FileItems, and it is when we receive a FileItem. In that case, we will return the same FileItem but with the `data` object updated (either a single package or an object with a package/release/record at the top level). So no need for the `root_path_max_length` logic, but the middleware can still yield FileItems.

jpmckinney: 👍 Sounds good

kingfisher_scrapy/spidermiddlewares.py
yolile: I think I covered this with the tests I've added.

jpmckinney:
I added a commit. I'm not sure what the appropriate logic should be for the last test case I added: https://github.com/open-contracting/kingfisher-collect/pull/944/files#diff-1d540996fd6e3d41c94656e00cd66d49ee9e5951ca8fc23a00a500475958a498R428-R431

The PR before my commit would not yield anything if the `root_path` pointed to an empty object. However, the old code would yield an object (like after my commit).

yolile:
> I'm not sure what the appropriate logic should be for the last test case 

Hmm, for data use, I guess it is better to drop/not yield empty objects. However, for checking the data and informing the publisher, I think leaving the data as it comes is preferred.

jpmckinney:
Yes - and hopefully it's rarely or never the case that it is empty, so shouldn't be too much a problem for users.

jpmckinney (approved):

jpmckinney:
FYI, I added a commit to not yield anything if a root_path points to an empty array, to simplify the logic 862dd0b7

PR #952, opened 2022-07-29
india_civic_data_lab: fix semi-colon-separated string

### LINKED ISSUE #888: Unflatten produces extensions as semi-colon-separated string, instead of list
For example, with the file from CivicDataLab at https://github.com/CivicDataLab/himachal-pradesh-health-procurement-OCDS/

Ravf95:
Closes #888

### DIFF
--- a/kingfisher_scrapy/spiders/india_civic_data_lab.py
+++ b/kingfisher_scrapy/spiders/india_civic_data_lab.py
@@ -18,7 +18,8 @@ class IndiaCivicDataLab(SimpleSpider):
     unflatten = True
     unflatten_args = {
         'metatab_name': 'Meta',
-        'metatab_vertical_orientation': True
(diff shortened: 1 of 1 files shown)


jpmckinney (approved):
I considered whether we should make this change in `class Unflatten`, but I guess it's possible for other future spiders to use a different schema (e.g. 1.0 vs 1.1), so this is fine.

yolile:
> but I guess it's possible for other future spiders to use a different schema (e.g. 1.0 vs 1.1), so this is fine.

Well, we set `ocds_version` in `BaseSpider` so I guess `Unflatten` could check for that attribute and set the `meta_schema` accordingly

jpmckinney:
Yes, that's fine too. I don't have a strong opinion either way.

PR #978, opened 2022-11-22
mexico_plataforma_digital_nacional: replace with data from Aguascalientes only

### LINKED ISSUE #976: Update mexico_plataforma_digital_nacional spider
https://www.plataformadigitalnacional.org/contrataciones used to have a link to a Google Drive JSON file. Now they point directly to https://datos.gob.mx/busca/dataset/concentrado-de-contrataciones-abiertas-de-la-apf. However, they are also publishing data from other data source now "Secretaría Ejecutiva del Sistema Estatal Anticorrupción de Aguascalientes". 

And now, they have an underlying API, for example: 

curl 'https://api.plataformadigitalnacional.org/s6/api/v1/search?supplier_id=SHCP' \
  -X 'POST'

To get the list of releases by publisher, so we could update the spider to use that API endpoint instead of the Google Drive file.

The only thing to decide is if we want two spiders, one per publisher (e.g., one for SHCP and another one for Aguascalientes) or just one for Plataforma Digital Nacional as of now.

yolile:
closes #976


yolile:
@jpmckinney, I've overridden the `parse_list` method from `IndexSpider` in this class as this endpoint requires a POST and a payload. I thought about incorporating this option as part of the `IndexSpider` base class itself, but maybe we can wait and see if there is another spider that requires this feature, as that class is already complex and with a lot of options, so I don't want to make it more complicated for just one spider.

kingfisher_scrapy/spiders/mexico_aguascalientes_sesea_plataforma_digital_nacional.py
```diff
@@ -0,0 +1,46 @@
...
+    use_page = True
+
+    # Local
+    url = 'https://api.plataformadigitalnacional.org/s6/api/v1/search?supplier_id=SESEA_AGS'
+
+    def start_requests(self):
+        yield scrapy.Request(self.url,
+                             meta={'file_name': 'page-0.json'},
+                             callback=self.parse_list,
+                             method='POST')
```
jpmckinney, line 37: Or do this indentation style:

```
some_thing = (
    a,
    b,
    c,
)
```

but never:

```
some_really_long_thing = (a,
                          b,
                          c)
```

```suggestion
        yield scrapy.Request(self.url, meta={'file_name': 'page-0.json'}, callback=self.parse_list, method='POST')
```

jpmckinney (commented):
Yes, I agree. We'd have to change build_request as well, which is probably too much.

kingfisher_scrapy/spiders/mexico_aguascalientes_sesea_plataforma_digital_nacional.py
yolile: Sorry, my IDE likes the second format!

jpmckinney: Ah, maybe it can be configured to use Black? https://marcobelo.medium.com/setting-up-python-black-on-visual-studio-code-5318eba4cd00 We haven't switched Kingfisher Collect to Black. Maybe there is another way to override the code style format.

jpmckinney (approved):

kingfisher_scrapy/spiders/mexico_aguascalientes_sesea_plataforma_digital_nacional.py
yolile: I will check that, thanks!

PR #1026, opened 2023-09-19
fix(spiders): use date filters as part of file names

### LINKED ISSUE #1025: Incremental updates with database store
I think there's a bug with the incremental update behaviour of the DatabaseStore.

If I understand correctly, crawl_time has to be set to the same value each time the spider is run to have incremental crawl.

The first time, 

1. it crawls
2. it saves data to the crawl_time directory in file names like `start.json`, `PageNumber-2.json`, ...
3. it creates one CSV file from all the crawl files
4. it creates the OCDS data table and inserts the data from the CSV file

Subsequent times,

1. it gets the latest publish date from the data in the table
2. it crawls from that publish date
3. 2. it saves data to the crawl_time directory in file names like `start.json`, `PageNumber-2.json`, ... **(I think it's overwriting files here)**
4. it creates one CSV file from all the crawl files
5. deletes the existing data and inserts the data from the CSV file

expected: All the data crawled previously plus the new data should be in the database
actual: Data in the overwritten files is missing from the database

Am I doing something wrong or is the overwriting an issue here? If I change crawl time for each crawl, none of the first crawl's data is included.

Some options I see:

- parse out the latest PageNumber- index and save to the next number (decouple the API page number from the saved file)
- use the existing data as part of the data going into compiling releases. That only solves the problem for people who enable compiling releases, but I want that, so it's fine for me.

yolile:
closes #1025

### DIFF
--- a/kingfisher_scrapy/spiders/colombia_api.py
+++ b/kingfisher_scrapy/spiders/colombia_api.py
@@ -32,7 +32,9 @@ class ColombiaAPI(LinksSpider):
     formatter = staticmethod(parameters('_id'))
 
     def start_requests(self):
+        from_date = self.from_date.strftime(self.date_format)
+        until_date = self.until_date.strftime(self.date_format)
         url = 'https://apiocds.colombiacompra.gov.co/apiCCE2.0/rest/releases/dates/' \
-              f'{self.from_date.strftime(self.date_format)}/{self.until_date.strftime(self.date_format)}'
+              f'{from_date}/{until_date}'
 
-        yield scrapy.Request(url, meta={'file_name': 'page-1.json'})
+        yield scrapy.Request(url, meta={'file_name': f'{from_date}-{until_date}-start.json'})
--- a/kingfisher_scrapy/spiders/colombia_api.py
+++ b/kingfisher_scrapy/spiders/colombia_api.py
@@ -36,4 +36,4 @@ class ColombiaAPI(LinksSpider):
         until_date = self.until_date.strftime(self.date_format)
         url = f'https://apiocds.colombiacompra.gov.co/apiCCE2.0/rest/releases/dates/{from_date}/{until_date}'
 
-        yield scrapy.Request(url, meta={'file_name': f'{from_date}-{until_date}-start.json'})
+        yield scrapy.Request(url, meta={'file_name': f'{from_date}.json'})
--- a/kingfisher_scrapy/spiders/colombia_api.py
+++ b/kingfisher_scrapy/spiders/colombia_api.py
@@ -34,7 +34,6 @@ class ColombiaAPI(LinksSpider):
     def start_requests(self):
         from_date = self.from_date.strftime(self.date_format)
         until_date = self.until_date.strftime(self.date_format)
-        url = 'https://apiocds.colombiacompra.gov.co/apiCCE2.0/rest/releases/dates/' \
-              f'{from_date}/{until_date}'
+        url = f'https://apiocds.colombiacompra.gov.co/apiCCE2.0/rest/releases/dates/{from_date}/{until_date}'
 
         yield scrapy.Request(url, meta={'file_name': f'{from_date}-{until_date}-start.json'})
--- a/kingfisher_scrapy/spiders/kyrgyzstan.py
+++ b/kingfisher_scrapy/spiders/kyrgyzstan.py
@@ -36,6 +36,7 @@ class Kyrgyzstan(LinksSpider):
             from_date = self.from_date.strftime(self.date_format)
             # The API requires the timezone and seconds in the since parameter.
             url = f'{url}?since={from_date}.00%2B06:00'
+            self.formatter = staticmethod(parameters('offset', 'since'))
         else:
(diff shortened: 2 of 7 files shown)


kingfisher_scrapy/spiders/kyrgyzstan.py
```diff
@@ -32,7 +32,10 @@ class Kyrgyzstan(LinksSpider):
 
     def start_requests(self):
```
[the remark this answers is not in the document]
yolile: The API is not working anymore, so hard to tell. But I will add `since`.

kingfisher_scrapy/spiders/kyrgyzstan.py
```diff
@@ -32,10 +32,10 @@ class Kyrgyzstan(LinksSpider):
...
         url = 'http://ocds.zakupki.gov.kg/api/tendering'
-        file_name = 'start.json'
         if self.from_date:
             from_date = self.from_date.strftime(self.date_format)
             # The API requires the timezone and seconds in the since parameter.
             url = f'{url}?since={from_date}.00%2B06:00'
-            file_name = f'{from_date}-{file_name}'
-        yield scrapy.Request(url, meta={'file_name': file_name})
+        else:
+            from_date = '1970-01-01T00:00:00'
```
yolile, line 40: @jpmckinney don't you think this will be confusing?

kingfisher_scrapy/spiders/colombia_api.py
```diff
@@ -32,7 +32,9 @@ class ColombiaAPI(LinksSpider):
...
 
     def start_requests(self):
+        from_date = self.from_date.strftime(self.date_format)
+        until_date = self.until_date.strftime(self.date_format)
         url = 'https://apiocds.colombiacompra.gov.co/apiCCE2.0/rest/releases/dates/' \
-              f'{self.from_date.strftime(self.date_format)}/{self.until_date.strftime(self.date_format)}'
+              f'{from_date}/{until_date}'
 
-        yield scrapy.Request(url, meta={'file_name': 'page-1.json'})
+        yield scrapy.Request(url, meta={'file_name': f'{from_date}-{until_date}-start.json'})
```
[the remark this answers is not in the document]
yolile: This one doesn't return the data in reverse chronological order.

kingfisher_scrapy/spiders/kyrgyzstan.py
jpmckinney: What is confusing? This spider, unlike others, doesn't set default_from_date, which I would have used here. In any case, using the UNIX epoch will have the same effect as a more recent date.

kingfisher_scrapy/spiders/colombia_api.py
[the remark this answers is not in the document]
jpmckinney: Ah, then use `from_date` and change the comment. I looked at the downloaded data on the registry, but I see now that the `date` is old (it's only the `publishedDate` that is recent).

kingfisher_scrapy/spiders/kyrgyzstan.py
yolile: The API is not working anymore anyway, so it really doesn't matter.

kingfisher_scrapy/spiders/portugal_base.py
```diff
@@ -20,10 +20,13 @@ class PortugalBase(LinksSpider):
...
         url = self.start_url
         if self.from_date and self.until_date:
-            url = f'{url}?contractStartDate={self.from_date.strftime(self.date_format)}' \
-                  f'&contractEndDate={self.until_date.strftime(self.date_format)}'
-
-        yield scrapy.Request(url, meta={'file_name': 'offset-1.json'})
+            from_date = self.from_date.strftime(self.date_format)
+            until_date = self.until_date.strftime(self.date_format)
+            url = f'{url}?contractStartDate={from_date}&contractEndDate={until_date}'
+            self.formatter = staticmethod(parameters('offset',  'contractStartDate', 'contractEndDate'))
```
jpmckinney, line 26:
```suggestion
            self.formatter = staticmethod(parameters('offset',  'contractStartDate'))
```

kingfisher_scrapy/spiders/united_kingdom_contracts_finder_base.py
```diff
@@ -32,8 +31,10 @@ def start_requests(self):
             from_date = self.from_date.strftime(self.date_format)
             until_date = self.until_date.strftime(self.date_format)
             url = f'{url}&publishedFrom={from_date}&publishedTo={until_date}'
+        else:
+            until_date = self.get_default_until_date(self).strftime(self.date_format)
 
-        yield scrapy.Request(url, meta={'file_name': 'page-1.json'}, callback=self.parse_page)
+        yield scrapy.Request(url, meta={'file_name': f'{until_date}.json'})  # reverse chronological order
```
jpmckinney, line 38: callback=self.parse_page got deleted?

jpmckinney (commented):
Two comments. Happy to merge once resolved.

PR #1027, opened 2023-09-26
feat(FilesStore): generate directory trees

### LINKED ISSUE #740: Some spiders create millions of files in a directory
This creates performance issues on some filesystems. The approach taken by caching libraries is to generate and fill in arbitrary directory trees; if we take the same approach, there is existing code/libraries to reuse. Otherwise, we can consider other ways to segment files, which might be specific to those spiders that have a lot of files.

yolile:
Closes #740

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,10 +30,12 @@ By default, these files are written to a ``data`` directory (you can :ref:`chang
    └── data
        └── zambia
            └── 20200102_030405
-               ├── <...>.json
-               └── <...>
+               └── C8E
+               └── D1D
+                   └── <...>.json
+                   └── <...>
 
-As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05). The crawl directory contains ``.json`` files – the OCDS data.
+As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05), then a subdirectory is created by creating a 3-characters long hash for each scraped file name. The crawl directory contains ``.json`` files – the OCDS data.
 
 .. toctree::
    :caption: Contents
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,12 +30,14 @@ By default, these files are written to a ``data`` directory (you can :ref:`chang
    └── data
        └── zambia
            └── 20200102_030405
-               └── C8E
+               ├── C8E
+               │   ├── <...>.json
+               │   └── <...>
                └── D1D
-                   └── <...>.json
+                   ├── <...>.json
                    └── <...>
 
-As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05), then a subdirectory is created by creating a 3-characters long hash for each scraped file name, ``D1D`` and ``C8E`` in this case. The crawl directory contains ``.json`` files – the OCDS data.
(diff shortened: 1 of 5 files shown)


jpmckinney (commented):
Need to update docs.

I updated the tests, where it's better to be explicit about expected values.

yolile:
> I updated the tests, where it's better to be explicit about expected values.

Yes, I wasn't sure which option was better so I implemented both, but yes, agree that explicit is better in tests

docs/index.rst
```diff
@@ -30,10 +30,12 @@ By default, these files are written to a ``data`` directory (you can :ref:`chang
...
            └── 20200102_030405
-               ├── <...>.json
-               └── <...>
+               └── C8E
+               └── D1D
+                   └── <...>.json
+                   └── <...>
 
-As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05). The crawl directory contains ``.json`` files – the OCDS data.
+As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05), then a subdirectory is created by creating a 3-characters long hash for each scraped file name, ``D1D`` and ``C8E`` in this case. The crawl directory contains ``.json`` files – the OCDS data.
```
jpmckinney, line 38:
```suggestion
As you can see, the ``data`` directory contains a ``zambia`` spider directory (matching the spider's name), which in turn contains a ``20200102_030405`` crawl directory (matching the time at which you started the crawl – in this case, 2020-01-02 03:04:05). Within the crawl directory,  ``.json`` files – the OCDS data – are split among subdirectories with opaque names, to not exceed filesystem limits.
```

docs/index.rst
```diff
@@ -30,10 +30,12 @@ By default, these files are written to a ``data`` directory (you can :ref:`chang
    └── data
        └── zambia
            └── 20200102_030405
-               ├── <...>.json
-               └── <...>
+               └── C8E
+               └── D1D
+                   └── <...>.json
+                   └── <...>
```
jpmckinney, line 36:
```suggestion
               ├── C8E
               │   ├── <...>.json
               │   └── <...>
               └── D1D
                   ├── <...>.json
                   └── <...>
```

kingfisher_scrapy/extensions/files_store.py
```diff
@@ -66,7 +67,8 @@ def spider_closed(self, spider, reason):
 
     def item_scraped(self, item, spider):
         """
-        If the item is a File or FileItem, writes its data to the filename in the crawl's directory.
+        If the item is a File or FileItem, writes its data to the filename in the crawl's directory, under a hashed
+        subdirectory using the filename.
```
jpmckinney, line 71:
```suggestion
        If the item is a File or FileItem, writes its data to the filename in a subdirectory of the crawl directory.
```

jpmckinney (dismissed):
Approve with a few copy-edits.

It occurs to me now that for our incremental crawls, we will need to either start from scratch, or write a little script to move files into appropriate directories.

I think it's better to write the script. Can you do that?

yolile:
> I think it's better to write the script. Can you do that?

Sure, I think it could be:

```` python
import os
import shutil
import zlib

directory = '/home/incremental/data/dominican_republic_api/20230713_000000'


def get_subdirectory(file_name):
    checksum = zlib.adler32(file_name.encode())
    checksum, dir_1 = divmod(checksum, 0x1000)
    return "%03X" % dir_1


for root, dirs, files in os.walk(directory):
    for name in files:
        if name.endswith('.json'):
            old_path = os.path.join(root, name)
            new_path = os.path.join(directory, get_subdirectory(name))
            os.makedirs(new_path, exist_ok=True)
            shutil.copy2(old_path, os.path.join(new_path, name))
            os.unlink(old_path)
````

jpmckinney:
That should work. Maybe os.rename instead of copy2 and unlink?

yolile:
The script was run.

PR #1052, opened 2024-01-26
periodic_spider: support date range and update dominican_republic_api to use it

### LINKED ISSUE #1051: dominican_republic_api: implement from and until date with new endpoint
Now they have https://api.dgcp.gob.do/api/docs/#/Release/get_api_date__date1___date2___page_, e.g. https://api.dgcp.gob.do/api/date/2023-01-01/2023-01-13/1

where we can query by a start and end date,  yyyy-mm-dd format.

yolile:
closes #1051

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           TEST_API_URL: http://localhost:${{ job.services.httpbin.ports[80] }}/anything/
         # For requests.post() in KingfisherProcessAPI2._post_synchronous().
         run: pytest -W error -W ignore::ResourceWarning -rs --cov kingfisher_scrapy
+      - run: python test_delayed_request_middleware.py
       - env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: coveralls --service=github
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,7 @@ jobs:
           RABBIT_URL: amqp://127.0.0.1:${{ job.services.rabbitmq.ports[5672] }}
           TEST_API_URL: http://localhost:${{ job.services.httpbin.ports[80] }}/anything/
         # For requests.post() in KingfisherProcessAPI2._post_synchronous().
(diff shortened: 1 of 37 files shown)


jpmckinney:
I added more tests and found an edge case. Not sure why we removed a second. It's from https://github.com/open-contracting/kingfisher-collect/commit/3494079760e1a58a69b57246bce37bcf3479a0c3

jpmckinney (approved):

jpmckinney:
Actually - I'm not 100% sure whether the change to `date_range_by_interval` returns the correct results for all the contexts it's called from (panama_dgcp_bulk, paraguay_dncp_base, periodic_spider),

It depends whether the "end" time is inclusive or exclusive, e.g. do we want pairs of days like Jan 1 to Jan 2, Jan 2 to Jan 3, etc. (end is exclusive) or Jan 1 to Jan 1, Jan 2 to Jan 2, etc. (end is inclusive).

paraguay_dncp_base is different, in that it formats the time as a full datetime (with seconds). The old logic treated the end time as inclusive (e.g. 23:59:59). With my update, it's now 00:00:00 on the next day.

We probably want an `inclusive` keyword argument to `date_range_by_interval` to return to correct values.

yolile:
> I added more tests and found an edge case. Not sure why we removed a second. It's from https://github.com/open-contracting/kingfisher-collect/commit/3494079760e1a58a69b57246bce37bcf3479a0c3

Hmm? That commit only changed the Paraguay spider, not `date_range_by_interval`, or are you referring to the Paraguay spider?

> We probably want an inclusive keyword argument to date_range_by_interval to return to correct values.

Sounds good, I will implement that

jpmckinney:
> Hmm? That commit only changed the Paraguay spider, not date_range_by_interval, or are you referring to the Paraguay spider?

That commit was Paraguay specific, but when we extracted that code into `date_range_by_interval`, the new `date_range_by_interval` function copied that logic.

yolile:
@jpmckinney could you please review it again? thank you!

jpmckinney (approved):

PR #1062, opened 2024-04-04
feat: add RetryableError logic in moldova spider

### LINKED ISSUE #1056: moldova: Inherit from BaseSpider and re-implement parse method
This logic:

```python
        data = response.json()

        if data.get('name') == 'Error':
            data['http_code'] = response.status
            yield self.build_file_error_from_response(response, errors=data)
            return
```

Needs to not only be on `parse_list` but `parse` as well.

It caused 35k errors in Sentry.

----

UnknownFormatError: top-level JSON value is a non-OCDS object

Sentry Issue: [REGISTRY-KINGFISHER-PROCESS-9C](https://open-contracting-partnership.sentry.io/issues/4940314221/?referrer=github_integration)

```
UnknownFormatError: top-level JSON value is a non-OCDS object
  File "process/management/commands/file_worker.py", line 81, in callback
  File "process/management/commands/file_worker.py", line 122, in process_file
    data_type = _get_data_type(collection_file)
  File "process/management/commands/file_worker.py", line 157, in _get_data_type
    detected_format, is_concatenated, is_array = detect_format(collection_file.filename)

Source moldova yields an unknown or unsupported format, skipping
```

yolile:
closes #1056

### DIFF
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -26,5 +26,5 @@ class UnknownArchiveFormatError(KingfisherScrapyError):
     """Raised when the archive format of a file can't be determined from the filename, from a spider callback"""
 
 
-class RetryableBadResponseError(KingfisherScrapyError):
-    """Raised when a URL returns a 200 HTTP response with a non-OCDS JSON object that could be retried"""
+class RetryableError(KingfisherScrapyError):
+    """Raised when the response is an error, but the request can be retried"""
--- a/kingfisher_scrapy/exceptions.py
+++ b/kingfisher_scrapy/exceptions.py
@@ -24,3 +24,7 @@ class MissingNextLinkError(KingfisherScrapyError):
 
 class UnknownArchiveFormatError(KingfisherScrapyError):
     """Raised when the archive format of a file can't be determined from the filename, from a spider callback"""
+
+
+class RetryableBadResponseError(KingfisherScrapyError):
+    """Raised when a URL returns a 200 HTTP response with a non-OCDS JSON object that could be retried"""
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -6,6 +6,7 @@ from zipfile import BadZipFile
 import ijson
 
 from kingfisher_scrapy import util
+from kingfisher_scrapy.exceptions import RetryableBadResponseError
 from kingfisher_scrapy.items import File, FileItem
 
 MAX_GROUP_SIZE = 100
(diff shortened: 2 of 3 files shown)


kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -269,13 +270,14 @@ async def process_spider_output(self, response, result, spider):
 
 class RetryDataErrorMiddleware:
```
jpmckinney, line 271: I wonder if maybe we don't need this middleware and can instead use Scrapy's existing https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.retry ?

I think maybe we just add zipfile.BadZipFile and kingfisher_scrapy.exceptions.RetriableError to [RETRY_EXCEPTIONS](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#retry-exceptions) (change its heredoc to be more generic – as we can later reuse this exception for other requests to retry)

kingfisher_scrapy/spiders/moldova.py
```diff
@@ -71,9 +72,7 @@ def parse_list(self, response):
         #   "code": "EHOSTUNREACH"
         # }
         if data.get('name') == 'Error':
-            data['http_code'] = response.status
-            yield self.build_file_error_from_response(response, errors=data)
-            return
+            raise RetryableBadResponseError
```
jpmckinney, line 75: Maybe extract the logic to a `self.raise_for_status(data)` method (inspired by requests' `response.raise_for_status()`).

kingfisher_scrapy/spidermiddlewares.py
yolile: Hmm, it seems that exceptions that are raised by spiders are not handled by this middleware https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#scrapy.downloadermiddlewares.DownloaderMiddleware.process_exception (I also tried with `moldova` that happens to be returning only errors now and the exception is just raised)

jpmckinney: Aha! Okay, so I'll review as-is.

kingfisher_scrapy/exceptions.py
```diff
@@ -24,3 +24,7 @@ class MissingNextLinkError(KingfisherScrapyError):
 
 class UnknownArchiveFormatError(KingfisherScrapyError):
     """Raised when the archive format of a file can't be determined from the filename, from a spider callback"""
+
+
+class RetryableBadResponseError(KingfisherScrapyError):
+    """Raised when a URL returns a 200 HTTP response with a non-OCDS JSON object that could be retried"""
```
jpmckinney, line 30:
```suggestion
class RetryableBadResponseError(KingfisherScrapyError):
    """Raised when the response is an error, but the request can be retried"""
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -269,13 +270,14 @@ async def process_spider_output(self, response, result, spider):
...
     """
-    Retries a request for a ZIP file up to 3 times, on the assumption that, if the spider raises a ``BadZipFile``
-    exception, then the response was truncated.
+    Retries a request up to 3 times. Either when the spider raises a ``BadZipFile`` exception, on the assumption that
+    the response was truncated, or when the spider raises a ``RetryableBadResponseError`` exception on the assumption
+    that the error was temporary.
```
jpmckinney, line 275: No need to assume in second case, as the exception is raised deliberately.

```suggestion
    Retries a request up to 3 times. Either when the spider raises a ``BadZipFile`` exception, on the assumption that
    the response was truncated, or when the spider raises a ``RetryableBadResponseError`` exception.
```

jpmckinney (approved):

PR #1063, opened 2024-04-05
fix(mexico_quien_es_quien*): remove records and fix releases

### LINKED ISSUE #1054: mexico_quien_es_quien_records: Downloads the same file over 2000 times
Can check this with:

```
md5sum /data/storage/kingfisher-collect/mexico_quien_es_quien_records/20240207_000156/**/*
```

----

IntegrityError: duplicate key value violates unique constraint "unique_record_identifiers"

Sentry Issue: [REGISTRY-KINGFISHER-PROCESS-90](https://open-contracting-partnership.sentry.io/issues/4529869296/?referrer=github_integration)

```
UniqueViolation: duplicate key value violates unique constraint "unique_record_identifiers"
DETAIL:  Key (collection_id, ocid)=(1984, ocds-0ud2q6-AA-019GYR069-E374-2017) already exists.

  File "django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)

IntegrityError: duplicate key value violates unique constraint "unique_record_identifiers"
DETAIL:  Key (collection_id, ocid)=(1984, ocds-0ud2q6-AA-019GYR069-E374-2017) already exists.

(11 additional frame(s) were not displayed)
...
  File "process/management/commands/file_worker.py", line 81, in callback
  File "process/management/commands/file_worker.py", line 136, in process_file
    _store_data(collection_file, package, releases_or_records, data_type, upgrade=False)
  File "process/management/commands/file_worker.py", line 276, in _store_data
    ).save()

IntegrityError maybe caused by duplicate message "{\"collection_id\":1984,\"collection_file_id\":248670680}", skipping
```

yolile:
records: pagination no longer working
releases: add dates filters to get all data

closes #1054


kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
```diff
@@ -1,22 +1,39 @@
...
 
     # IndexSpider
-    base_url = 'https://api.quienesquien.wiki/v3/contracts?sort=date&sort_direction=desc'
+    result_count_pointer = '/count'
+    limit = 1000
+
+    # PeriodicSpider
+    formatter = staticmethod(parameters('start_date_min', 'start_date_max', 'offset'))
+    pattern = 'https://api.quienesquien.wiki/v3/contracts?start_date_min={0:%Y-%m-%d}&start_date_max={' \
+              '1:%Y-%m-%d}&limit=25&offset=0 '
```
jpmckinney, line 38: Why are we setting `limit` to 25? I think we can also omit `offset`. (This can then fit on one line.)

kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
```diff
@@ -1,22 +1,39 @@
...
       https://qqwapi-elastic.readthedocs.io/es/latest/
     Swagger API documentation
       https://api.quienesquien.wiki/v3/docs/
     """
     name = 'mexico_quien_es_quien_releases'
+    download_delay = 1
 
     # BaseSpider
+    default_from_date = '1999-01-01'
+    default_until_date = '2021-12-31'
```
jpmckinney, line 24: They stopped making updates?

kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
```diff
@@ -1,22 +1,39 @@
...
+    date_format = 'date'
```
jpmckinney, line 25: I assume they set their dates correctly (some publications create huge numbers of releases on specific days, as the date is set according to when a software process runs, rather than based on the data)

jpmckinney (commented):
Add removal to history.md as well (need to make a new list, as reason is different).

kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
yolile: Ah good catch

kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
yolile: Yes. I've tried and the last date with results is 2021-12

kingfisher_scrapy/spiders/mexico_quien_es_quien_releases.py
yolile: We cannot omit offset because we use it in

`formatter = staticmethod(parameters('start_date_min', 'start_date_max', 'offset'))`

yolile: Actually, we also need to set the limit (to `limit`) as the default is 25

jpmckinney: Doesn't IndexSpider set the limit and offset? I guess not for the first request. If that's the reason, we should still have `limit=1000` for consistency.

jpmckinney (approved):

PR #1066, opened 2024-04-10
1058 filter invalid json files

### LINKED ISSUE #645: pakistan_ppra_api: Invalid JSON
https://www.ppra.org.pk/api/index.php/api/record/442076 contains unescaped quote characters like:

> "description": "1. Corri pub on 29 Jan 21 against tender no. 16-0982-00 qty is amended"FOR" 216978 "READ" 127540. 2. Corri pub on 04 Feb 21 against tender no. 07-1317-5-00 Tender Delivery Period and opening date are amended. 3. Corri pub on 8 Feb 21 against tender no 16-0982-00 tender opening date is amended "FOR" 8 Feb 21 "READ"  1 Mar 21. 4. Corri pub on 12 Feb 21 against tender no. 02-1619-00 tender opening date is amended "FOR" 12 Feb 21 "READ" 19 Feb 21. 5. Corri pub on 12 Feb 21 against tender no. 01-1438-00 tender opening date is amended "FOR" 12 Feb 21 "READ" 23 Feb 21. 6. Corri pub on 18 Feb 21 against tender no. 02-1619-00 tender opening date is amended "FOR" 19 Feb 21 "READ" 26 Feb 21. 7. Corri pub on 24 Feb 21 against tender no. 07-1317-00 tender opening date is amended "FOR" 22 Feb 21 "READ" 02 Mar 21. 8. Corri pub on 26 Feb 21 against tender no. 16-0982-00 tender opening date is amended "FOR" 01 Mar 21 "READ" 15 Mar 21 . See more detail in attach file, please.",

It seems to all be `"FOR"` and `"READ"`, so we can maybe escape those quote characters and see if the rest works.

### LINKED ISSUE #876: costa_rica_poder_judicial_releases: Invalid JSON (2023)
Download https://pjcrdatosabiertos.blob.core.windows.net/datosabiertos/OpenContracting/ocds-fnha3a-FULL.zip

Open, for example, 2021/001261-2021.json

The JSON is invalid:

```
"planning":{"requisition":[{"id":"004435-SR-2021","rationale":","value":{"amount":478500.00,"currency":"CRC"}}]
```

Note that rationale has no closing `"`.

### LINKED ISSUE #886: croatia: Invalid JSON
croatia gets creative with JSON like `[{"id":"1190488","awardID"}]` seen in `list-EOJN_OCD_01_02_2019.json`.

Please report the issue to the publisher in January.

### LINKED ISSUE #957: `nigeria_anambra_state_records` invalid JSON
Scrapes return 0 rows, attempting to use the URL via the browser (Firefox) results in "Deceptive site warnings". Kingfisher Collect logs indicate DNS issues.

### LINKED ISSUE #963: nigeria_osun_state, nigeria_enugu_state returns incomplete JSONs
[Senty issue](https://sentry.io/organizations/open-contracting-partnership/issues/3616401875/?referrer=alert_email&alert_type=email&alert_timestamp=1663968890791&alert_rule_id=7248661&environment=production)

### LINKED ISSUE #964: indonesia_opentender yields invalid JSON
For `export-ocds-batch-year-2021-lpse-127-ocds-data-tender-24092022143442.json`

From https://opentender.net/api/tender/export-ocds-batch?year=2021&lpse=127

Sentry Issue: [DATA-REGISTRY-KINGFISHER-PROCESS-6Y](https://sentry.io/organizations/open-contracting-partnership/issues/3616401875/?referrer=github_integration)

```
IncompleteJSONError: parse error: trailing garbage
          rindustrian", "id": "K27"}}]}tity"]}, {"name": "PT. HUDA TAT
                     (right here) ------^

  File "process/management/commands/file_worker.py", line 57, in callback
  File "process/management/commands/file_worker.py", line 100, in process_file
    package, releases_or_records = _read_data_from_file(collection_file.filename, data_type)
  File "process/management/commands/file_worker.py", line 174, in _read_data_from_file
    for prefix, event, value in ijson.parse(ControlCodesFilter(f)):

Spider indonesia_opentender yields invalid JSON, skipping
```

### LINKED ISSUE #1036: panama_dgcp_records yields incomplete JSON
e.g. https://open-contracting-partnership.sentry.io/issues/4561712790/?project=1404981&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0

Not sure what proportion of files have this issue.

### LINKED ISSUE #1058: Filter out invalid and incomplete JSON
If we can filter these out, then we can include more publications in the registry where this issue occurs in a small subset of all available files.

Typically, filtering is done in the item pipeline. However, spider middlewares run prior to the item pipeline, and we parse the JSON in these middlewares. (In some cases, we parse the JSON in the spider, but only when we have to in order to create URLs.)

Maybe we mark the item with a first middleware, and the other middlewares are skipped if that mark is present. The pipeline could then drop these marked items, and log the total.

Related to #1055, we maybe want to set a threshold such that the spider closes with a different reason if the threshold for invalid JSON files is exceeded.

yolile:
closes #1058 
closes #1036 
closes #964 
closes #963 
closes #957 
closes #886 
closes #876 
closes #645


jpmckinney:
Did you do crawls for the individual spiders? I would be interested to know if the invalid JSON are a small or large proportion. (You can do a sample of 100 or something.)

jpmckinney:
From #964 
> With https://github.com/open-contracting/kingfisher-collect/commit/f8d47ca45b3af4b020d1c1f8a235cb4c98259c55 this problem didn't happen again. However, as the publisher is working on a better approach to generate the files, I will keep this issue open so we can remove the download limitations from the spider when the new solution is implemented.

I assume you are okay with closing the issue and leaving those limits for now.

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -290,3 +292,25 @@ def process_spider_exception(self, response, exception, spider):
...
+        :returns: a generator of File or FileItem objects, in which the ``data`` field is parsed JSON
+        """
+        async for item in result:
+            if not isinstance(item, (File, FileItem)) or not spider.validate_json:
+                yield item
+                continue
+
+            try:
+                json.loads(item.data)
+            except (JSONDecodeError, TypeError):
```
jpmckinney: In what scenario is TypeError raised?

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -75,6 +75,32 @@ async def process_spider_output(self, response, result, spider):
...
+            ):
+                yield item
+                continue
+
+            try:
+                json.loads(item.data)
+            except json.JSONDecodeError:
+                item.invalid_json = True
+
+            yield item
```
jpmckinney, line 101: What if instead of setting invalid_json, we just don't yield anything? In that case, we'll just want to use the spider's logger to log what we dropped (like in LogFormatter). This will avoid some repetitive logic elsewhere.

jpmckinney (commented):
* [ ] use the spider's logger to log what we dropped (like in LogFormatter)
* [x] I think we also wanted to update the spider's stats with the number of invalid JSON items.

kingfisher_scrapy/spidermiddlewares.py
jpmckinney: Scrapy's DepthMiddleware works this way, so I'll make this change. https://github.com/scrapy/scrapy/blob/48c5a8c98f545e35708a580e724d0b8e1ada5e6e/scrapy/spidermiddlewares/depth.py#L46-L58

jpmckinney (dismissed):

kingfisher_scrapy/spidermiddlewares.py
yolile: It happened with one test (empty dict) which you already covered now, so I guess it shouldn't happen in real life

yolile:
> Did you do crawls for the individual spiders? 

I tested it with two Nigerian states, one dropped 1 file, and the other 2. I could try with all the affected spiders if you want, or do it in the registry directly

yolile:
This approach fails when the data is compressed, as the middleware runs before the data is decompressed, e.g. with `croatia`.

kingfisher_scrapy/spidermiddlewares.py
jpmckinney: Aha, yeah, `json.loads({})` or `json.loads([])`. These now passthrough instead of failing.

jpmckinney:
> I tested it with two Nigerian states, one dropped 1 file, and the other 2. I could try with all the affected spiders if you want, or do it in the registry directly

Sure, please check the registry logs when the time comes.

> This approach fails when the data is compressed, as the middleware runs before the data is decompressed, e.g. with croatia.

Aha, for Croatia we had `ReadDataMiddleware` read the data (if it survived that long without being read). Fixed in d6e8a7fa (similar to AddPackageMiddleware)

jpmckinney:
We can add ValidateJSONMiddleware to the `test_bytes_or_file` test to cover this.

yolile:
> Aha, for Croatia we had ReadDataMiddleware read the data 

And also replace `item.data` with `data.read()`, otherwise, we get empty JSON files.

I tested locally the spiders that are small and faster enough with these results:

nigeria_anambra_state: 23/169 (13,6% missing) (one ocid per file)
nigeria_osun_state: 2/164 (1% missing) (one ocid per file)
nigeria_enugu_state: 1/7 (14% missing) (one year per file, 2020 invalid, but 2023 and 2024 without releases)
croatia: 9/65 (13,8% missing) (one month per file)
costa_rica_poder_judicial_releases: 1306/11991 (10% missing)  (one ocid per file)
pakistan_ppra_api is down
panama_dgcp_records: too slow/big

yolile:
@jpmckinney I've included the test, do you think we are ok to merge this now?

Issue #995, opened 2023-01-27
Consider using `@dataclass` in `items.py`


jpmckinney:
https://docs.scrapy.org/en/latest/topics/items.html

Once we deploy to Python >= 3.7.

jpmckinney:
I did a bit more testing and Pydantic v1 and v2 both have issues, if the field is allowed to be a dict (v1) or a list (v2) in addition to `Any`. If `dict` and `list` are not allowed, it's fine.

```python
import pydantic
import io
import zipfile
from pathlib import Path
from typing import Any


class AnyData(pydantic.BaseModel, arbitrary_types_allowed=True):
    data: Any


class DictOrAnyData(pydantic.BaseModel, arbitrary_types_allowed=True):
    data: dict | Any


class ListOrAnyData(pydantic.BaseModel, arbitrary_types_allowed=True):
    data: list | Any



path = Path("test.txt")
path.write_text("foo\nbar\nbaz\nbzz\nzzz")

t = AnyData(data=path.open("rb"))
print(t.data.read())

t = DictOrAnyData(data=path.open("rb"))
print(t.data.read())

t = ListOrAnyData(data=path.open("rb"))
print(t.data.read())
```

PR #1068, opened 2024-04-11
Use Pydantic

jpmckinney:
Builds on #1066 

#995 

Some differences:

* pydantic validates on instantiation (and we opt-in to also validate on assignment). This means the full item needs to be instantiated at once, instead of composed over several instructions. This currently works with how we do things. This mainly means a programming error will be raised early, rather than once an item reaches the pipeline.
* pydantic casts values, unless strict is set to true. We use `strict` where possible, but there is no strict option for dicts, so we add our own validator.
* There was a weird bug where, if a pydantic item is `return`ed instead of `yield`ed, then some code somewhere iterates over it, which turns it into tuples. Anyway, we should always yield.


jpmckinney:
Tests pass and I also ran:

```
scrapy crawlall --sample 1 --loglevel=ERROR
```

I added this to the crawall command to skip the ones with authentication and the many broken Digiwhist spiders:

```python
if spider_name.startswith(('openopps', 'paraguay')) or spider_name.endswith('digiwhist'):
    continue
```

yolile:
Ah, nice and clear!

yolile (approved):

PR #1069, opened 2024-04-11
feat: remove check for default steps in kigfisher process

### LINKED ISSUE #861: Change default to skip checks
Originally, Collect was used primarily to give feedback to publishers, in which case it made sense to check the data by default in Process. (Collect passes the desired steps to Process.)

Now, we often use Collect to perform analyses, where we don't care about or even look at the check results.

Since the check step is very slow, unnecessary checks from analysis collections interfere with the needed checks for feedback collections.

I propose that we change checks to "opt-in" so that analysts get their results quicker without extra effort.

@yolile Can you check with ODS and CDS if they agree?

cc @duncandewhurst as this relates to your Slack message.

yolile:
closes #861

### DIFF
--- a/kingfisher_scrapy/base_spiders/base_spider.py
+++ b/kingfisher_scrapy/base_spiders/base_spider.py
@@ -114,7 +114,8 @@ class BaseSpider(scrapy.Spider):
         self.kingfisher_process_note = note
         self.kingfisher_process_keep_collection_open = keep_collection_open == 'true'
         if steps is None:
-            self.kingfisher_process_steps = self.available_steps
+            # Set the compile step only by default, as the check step is very slow.
+            self.kingfisher_process_steps = {'compile'}
         else:
             self.kingfisher_process_steps = set(steps.split(',')) & self.available_steps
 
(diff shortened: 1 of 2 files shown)


kingfisher_scrapy/base_spiders/base_spider.py
```diff
@@ -114,7 +114,8 @@ def __init__(self, sample=None, path=None, from_date=None, until_date=None, craw
         self.kingfisher_process_note = note
         self.kingfisher_process_keep_collection_open = keep_collection_open == 'true'
         if steps is None:
-            self.kingfisher_process_steps = self.available_steps
+            # Set the compile step only by default, as the check step is very slow.
```
jpmckinney, line 117: I think we can omit the comment, as the reason is that it's both slow AND we don't always need it – but that's more about our own policies, than about the project.

jpmckinney (commented):
Just need to update test expectations, and one comment.

yolile:
> Just need to update test expectations

Does kingfisher process expect to always receive a dict with all the available steps with their boolean values? e.g. compile: True, check: false, etc?

jpmckinney:
No, if something is missing, it defaults to False.

jpmckinney (approved):

PR #1106, opened 2024-10-10
fix: parse_list_loader in chile_compra_api is a generator

yolile:
I'm not sure why, but `parse_list_loader` is a generator when `ChileCompraAPIBase` overrides it


jpmckinney:
I made an error in 5048b04d1. I think I copy-pasted from EuropeanDynamicsBase. See my PR: #1107

PR #1122, opened 2024-11-04
fix: use new endpoints for mexico_nuevo_leon_*

### LINKED ISSUE #1100: mexico_nuevo_leon_releases: update endpoints
https://smpu.nl.gob.mx/acceso/DatosAbiertos/JSONsInfraestructuraAbierta.rar linked from https://smpu.nl.gob.mx/transparencia/acerca-del-proyecto requires HTTP authentication (no login provided, as far as I can see).

yolile:
closes #1100

### DIFF
--- a/kingfisher_scrapy/spiders/mexico_nuevo_leon_records.py
+++ b/kingfisher_scrapy/spiders/mexico_nuevo_leon_records.py
@@ -1,23 +1,28 @@
-import scrapy
+from kingfisher_scrapy.base_spiders import PeriodicSpider
+from kingfisher_scrapy.util import browser_user_agent, parameters
 
-from kingfisher_scrapy.base_spiders import SimpleSpider
 
-
-class MexicoNuevoLeonRecords(SimpleSpider):
+class MexicoNuevoLeonRecords(PeriodicSpider):
     """
     Domain
       Secretaría de Movilidad y Planeación Urbana de Nuevo León
+    Spider arguments
+      from_date
+        Download only data from this year onward (YYYY format). Defaults to '2013'.
+      until_date
+        Download only data until this year (YYYY format). Defaults to the current year.
     Bulk download documentation
       https://smpu.nl.gob.mx/transparencia/publicaciones
     """
 
     name = 'mexico_nuevo_leon_records'
+    user_agent = browser_user_agent  # to avoid HTTP 403
 
     # SimpleSpider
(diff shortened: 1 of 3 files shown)


kingfisher_scrapy/spiders/mexico_nuevo_leon_records.py
```diff
@@ -1,23 +1,28 @@
...
-    def start_requests(self):
-        yield scrapy.Request(
-            'https://smpu.nl.gob.mx/siasi_ws/api/ocds/DescargarRecordPackage',
-            meta={'file_name': 'records.json'}
-        )
+    # PeriodicSpider
+    date_format = 'year'
+    pattern = 'https://smpu.nl.gob.mx/siasi_ws/api/ocds/ListarProduccionXAnio?anio=%5B%7B"value":"{0}"%7D%5D'
+    formatter = staticmethod(parameters('anio'))
+    default_from_date = '2013'
```
jpmckinney, line 28: date_format and default_from_date go in a `# BaseSpider` section: https://kingfisher-collect.readthedocs.io/en/latest/contributing/index.html

kingfisher_scrapy/spiders/mexico_nuevo_leon_releases.py
```diff
@@ -1,12 +1,16 @@
-import scrapy
+from kingfisher_scrapy.spiders.mexico_nuevo_leon_records import MexicoNuevoLeonRecords
+from kingfisher_scrapy.util import components
 
-from kingfisher_scrapy.base_spiders import CompressedFileSpider
 
-
-class MexicoNuevoLeonReleases(CompressedFileSpider):
+class MexicoNuevoLeonReleases(MexicoNuevoLeonRecords):
```
jpmckinney, line 5: Do we want this spider? I don't think we typically use a records endpoint to create a releases spider.

yolile: I guess we want it as it would be the only way to get the original releases, as the record contains compiled and linked releases only. I could create a base class instead of inheriting from Records though

jpmckinney: Aha, makes sense. Please create a base class, since in future we might not want changes to the records spider to cause changes in the releases spider.

jpmckinney (approved):

PR #1128, opened 2024-12-03
fix: root path and proper file names for mexico_nuevo_leon_infra_abierta_records

### LINKED ISSUE #1100: mexico_nuevo_leon_releases: update endpoints

yolile:
closes #1100

### DIFF
--- a/kingfisher_scrapy/spiders/mexico_nuevo_leon_infra_abierta_base.py
+++ b/kingfisher_scrapy/spiders/mexico_nuevo_leon_infra_abierta_base.py
@@ -1,5 +1,12 @@
+import json
+
 from kingfisher_scrapy.base_spiders import PeriodicSpider
 from kingfisher_scrapy.util import BROWSER_USER_AGENT, parameters
+
+
+def parser(value):
+    # The 'anio' query string parameter looks like [{"value":"2024"}] (before URL encoding).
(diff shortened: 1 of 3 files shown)


jpmckinney (commented):
slice_url is not a robust approach (even small differences in URL construction would cause it to break). Maybe instead add a function to the spider file that json.loads the anio query string parameter to read the value?

yolile:
> slice_url is not a robust approach (even small differences in URL construction would cause it to break). 

I thought the same, but we are also using the `pattern`  approach where the same applies, doesn't it?
In any case, I will try your suggestion instead

jpmckinney (approved):
I made a commit.

Putting logic in build_file_item is brittle, because build_file_item only gets called thanks to `root_path` being set. If NL stops wrapping records in arrays, the filename fix stops working.

I know that `pattern` limits the likelihood of a break, but it's not obvious to a future contributor that changing e.g. `ListarProduccionXAnio` to `ListarProduccionAnio` also requires changing integer offsets.

PR #1129, opened 2024-12-03
fix: italy_anac to handle bid files

### LINKED ISSUE #1127: italy_anac: Use ijson instead of response.json()
Some files are 2.5GB+. Reading the entire file into memory can cause `MemoryError` if the server is under load.

Like:

```
curl -I https://dati.anticorruzione.it/opendata/download/dataset/ocds/filesystem/bulk/2024/02.json
curl -I https://dati.anticorruzione.it/opendata/download/dataset/ocds/filesystem/bulk/2024/04.json
```

Alternatively (and maybe both better and easier)

1. Use the `BigFileSpider` base class, since these files are too big for most users anyway
2. Add support to ResizePackageMiddleware for a new `ocid_fallback`  class attribute, that is a function to generate the OCID

yolile:
closes #1127 
@jpmckinney checking if this was what you had in mind, before adding the documentation and tests

### DIFF
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -238,7 +238,7 @@ class ResizePackageMiddleware:
                 if sample_filled(spider, number):
                     return
 
-                # Kingfisher Process merges only releases with ocids.
+                # Kingfisher Process merges only releases and records with OCIDs.
                 if hasattr(spider, 'ocid_fallback'):
                     for entry in items:
                         if 'ocid' not in entry:
--- a/kingfisher_scrapy/spidermiddlewares.py
+++ b/kingfisher_scrapy/spidermiddlewares.py
@@ -214,7 +214,9 @@ class ResizePackageMiddleware:
     """
     If the spider's ``resize_package`` class attribute is ``True``, split the package into packages of 100 releases or
     records each. Otherwise, yield the original item.
-    Optionally, implement an ocid_fallback function that returns an alternative ocid value if the ocid is not set.
+
+    Optionally, implement an ``ocid_fallback`` method on the spider, which accepts a release (or record) and returns an
+    an ``ocid`` value, to be used if the ``ocid`` field is not set.
     """
 
     async def process_spider_output(self, response, result, spider):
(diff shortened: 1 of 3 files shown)


kingfisher_scrapy/spiders/italy_anac.py
```diff
@@ -22,6 +22,9 @@ class ItalyANAC(SimpleSpider):
     # SimpleSpider
     data_type = 'release_package'
 
+    # ResizePackageMiddleware
+    resize_package = True
+
```
jpmckinney, line 27: BigFileSpider sets this. You can delete it here.

kingfisher_scrapy/spiders/italy_anac.py
```diff
@@ -33,12 +36,8 @@ def parse_list(self, response):
...
-    @handle_http_error
-    def parse(self, response):
-        data = response.json()
-        for release in data['releases']:
-            # Kingfisher Process merges only releases with ocids. Extract the ocid from the release id as a fallback.
-            # For example: ocds-hu01ve-7608611 from ocds-hu01ve-7608611-01.
-            if 'ocid' not in release:
-                release['ocid'] = '-'.join(release['id'].split('-')[:3])
-        yield self.build_file_from_response(response, data_type=self.data_type, data=data)
+    @staticmethod
```
jpmckinney, line 39: I'm surprised `@staticmethod` works, since you call it with `spider.ocid_fallback`, which automatically provides `self`. (?)

kingfisher_scrapy/spiders/italy_anac.py
```diff
@@ -33,12 +36,8 @@ def parse_list(self, response):
...
+    def ocid_fallback(release):
+        # Kingfisher Process merges only releases with ocids. Extract the ocid from the release id as a fallback.
+        # For example: ocds-hu01ve-7608611 from ocds-hu01ve-7608611-01.
```
jpmckinney, line 42:
```suggestion
        # Extract the ocid from the release id as a fallback, like ocds-hu01ve-7608611 from ocds-hu01ve-7608611-01.
```

kingfisher_scrapy/spidermiddlewares.py
```diff
@@ -238,6 +238,12 @@ async def process_spider_output(self, response, result, spider):
                 if sample_filled(spider, number):
                     return
 
+                # Kingfisher Process merges only releases with ocids.
```
jpmckinney, line 241:
```suggestion
                # Kingfisher Process merges only releases and records with OCIDs.
```

jpmckinney (commented):
Okay to merge if suggestion accepted.

yolile:
Fixed by https://github.com/open-contracting/kingfisher-collect/commit/e087dfeaf98026eb9dc033d5f34bae15c7214912

PR #1141, opened 2025-01-07
fix(poland_digiwhist): The usual URL is 404

jpmckinney:


yolile:
Same for: TED, Slovenia and Czech Republic

PR #1166, opened 2025-06-09
feat: add special case for malformed JSONs in LinksSpider

### LINKED ISSUE #1163: LinksSpiders should have a special failure case
In the registry, we fail crawls that have an error rate of 15% or greater.

For LinksSpider, like `united_kingdom_fts`, it just takes one error for the crawl to stop, since it needs to paginate one page at a time and cannot guess the URL of the next page. In other words, as soon as it crawls 7 pages, it can fail, and the registry will consider it a pass (1/7 = 14%). This is of course a problem if we know that the full dataset contains hundreds or thousands of pages.

So, we should add a way for Collect to communicate to the Registry (via the Scrapy log) that the crawl was forced to stop, and that it should be considered incomplete.

Once this is implemented, we can create a corresponding issue in data-registry.

cc @yolile

yolile:
closes #1163

### DIFF
--- a/kingfisher_scrapy/base_spiders/links_spider.py
+++ b/kingfisher_scrapy/base_spiders/links_spider.py
@@ -1,4 +1,4 @@
-from json import JSONDecodeError
+import json
 
 from jsonpointer import resolve_pointer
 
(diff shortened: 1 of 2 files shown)


jpmckinney (dismissed):
Yup! Next we can check for MissingNextLinkError errors in the registry, and fail the job if so.

PR #1300, opened 2026-07-23
feat(uganda_releases): implement new endpoint

### LINKED ISSUE #1249: uganda_releases: use new URL
https://cdn.ppda.go.ug/api/open-data/v2/ocds/download? is the URL for downloads now

yolile:
closes #1249

Note: the server is failing for all URLS (it seems they have an issue with their Redis), so I can't test if this is working as expected


kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,73 @@ class UgandaReleases(PeriodicSpider):
     """
 
     name = "uganda_releases"
-    # https://gpp.ppda.go.ug/public/open-data/ocds/ocds-datasets generates URLs with JavaScript. We increment
-    # the 'code' parameter until it 404s. As such, we can't disambiguate expected from unexpected 404s.
-    handle_httpstatus_list = [404]
-    # Returns HTTP 403 if too many requests. (1 is too short.)
+    # To avoid adding too many links to the queque.
```
jpmckinney: Typo. But also why is too many links in queue a problem?

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,73 @@ class UgandaReleases(PeriodicSpider):
...
+        else:
+            self.logger.error("Export request failed: %r", data)
 
-        yield self.build_request(
-            replace_parameters(response.request.url, code=int(get_parameter_value(response.request.url, "code")) + 1),
-            formatter=self.formatter,
-        )
+    def parse_status(self, response):
+        # Statuses that indicate the export job is not ready yet, so we keep polling.
+        # Currently, the server is failing for all URLs, so we don't know the success status code
```
jpmckinney: Let's remove this comment for now as it will be stale once it starts working

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,73 @@ class UgandaReleases(PeriodicSpider):
...
+        in_progress_statuses = frozenset({"queued", "processing", "pending", "running", "in_progress", "started"})
```
jpmckinney: We can promote this to an all-caps module-level type:

```python
from typing import Literal

IN_PROGRESS_STATUES = Literal["queued", "processing", ...]
```

and then use a `match`/`case` statement instead of if/else.

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,73 @@ class UgandaReleases(PeriodicSpider):
...
+            request = response.request.copy()
+            request.meta["retries"] = attempts
+            request.dont_filter = True
+            yield request
+        else:
+            yield self.build_request(
+                f"{self.url_prefix}exports/{meta['job_id']}/download",
+                formatter=None,
+                meta={"file_name": meta["file_name"]},
+                callback=self.parse,
```
jpmckinney: This is the default callback so we can omit this argument.

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,73 @@ class UgandaReleases(PeriodicSpider):
...
+        if data.get("success"):
+            yield scrapy.Request(
+                data["status_url"],
+                meta={
+                    "file_name": response.request.meta["file_name"],
+                    "job_id": data["job_id"],
+                    "wait_time": 30,
+                },
+                callback=self.parse_status,
+                dont_filter=True,
```
jpmckinney: Is it necessary for the initial request to set dont_filter?

kingfisher_scrapy/spiders/uganda_releases.py
yolile: It is just to avoid querying the status endpoint too many times (assuming the processing will take longer time if we request multiple files at once)

jpmckinney:
Ok, since the server isn't responding, I can't finish review. I find it a little surprising that the status field has 6 values for in-progress, for example. I'm also curious about the response passed that `parse_job` receives.

kingfisher_scrapy/spiders/uganda_releases.py
```diff
@@ -22,36 +25,69 @@ class UgandaReleases(PeriodicSpider):
...
-        yield self.build_request(
-            replace_parameters(response.request.url, code=int(get_parameter_value(response.request.url, "code")) + 1),
-            formatter=self.formatter,
-        )
+    def parse_status(self, response):
+        data = response.json()
+        meta = response.request.meta
+        match data.get("status"):
+            case "failed":
+                self.logger.error("Export job %s failed: %s", meta["job_id"], data.get("message"))
```
jpmckinney, line 77: For the two error messages, I think clearer to use the file_name, no?

yolile:
I've marked this as blocked until the server is working again

